PR #10, opened 2020-06-05
Add support for Electron pre-built binaries

xxuejie:


packages/indexer/package.json
```diff
@@ -32,10 +33,12 @@
   "scripts": {
     "build": "neon build",
     "build-release": "neon build --release",
-    "fmt": "prettier --write \"{lib,tests}/**/*.js\" index.d.ts && cd native && cargo fmt --all --",
-    "lint": "eslint -c ../../.eslintrc.js \"lib/**/*.js\"",
-    "install": "node-pre-gyp install --fallback-to-build=false || npm run build-release",
+    "build-for-electron": "electron-build-env --electron 9.0.2 -- neon build --release",
```
katat, line 36: Does this mean the pre-build binaries will only build for the electron version `9.0.2`?

If so, then it may still have the version mismatch issue, as the older versions of electron won't be able to use this native module.

If we can't provide all the pre-built binaries for all the different versions of electron, maybe we need to display a warning in the console to let the clients know they need to build it following the second approach documented in the readme. The other option would be to automatically build the binary based on the installed version of electron and display a notice for the clients to install the `cargo` tool if it is not available in the environment.

xxuejie: 2 points:

1. Even when we are specifying `9.0.2`, if we look at [release page](https://github.com/nervosnetwork/lumos/releases/tag/v0.2.5-rc.3), we can notice that this flow actually generates binaries for Electron 9.0, meaning all versions of electron 9.0.x should share the binary
2. Later when there are newer Electron releases, we can always come back here to revise the code, so we are building for more release versions other than just 9.0. It is unlikely we will support all versions, but we can at least support several major important releases.

PR #11, opened 2020-06-06
chore: Use yarn workspaces to replace lerna

xxuejie:


xxuejie:
@classicalliu @bitrocks When you have a moment please give this a try, we will merge this when everyone agrees with this new flow.

PR #12, opened 2020-06-07
feat: Add multisig script generation skeleton

classicalliu:
* I'm not sure should we use `R` and `M` or more readable `require_n` and `threshold`, former one is more friendly to who's familiar with `secp256k1_blake160_multisig`.
* Should we retain `fromAddress`, or we can delete this parameter and generate `fromScript` from `fromInfo(R, M, publicKeyHashes and so on...)` ?
* Should we expose `serializeMultisigScript` and `multisigArgs`, these will help to generate multisig address, or we can move to somewhere like `utils` ?

### DIFF
--- a/packages/common-scripts/index.d.ts
+++ b/packages/common-scripts/index.d.ts
@@ -1,5 +1,23 @@
 import { TransactionSkeleton } from "@ckb-lumos/types"
 
+export type Address = string
+export type Config = any // TODO: define this type later
+
+/**
+ * secp256k1_blake160_multisig script requires S, R, M, N and public key hashes
+ * S must be zero now
+ * and N equals to publicKeyHashes size
+ * so only need to provide R, M and public key hashes
+ */
+export interface MultisigScript {
+  /** first nth public keys must match, 1 byte */
+  R: number,
+  /** threshold, 1 byte */
+  M: number,
+  /** blake160 hashes of compressed public keys */
+  publicKeyHashes: string[],
+}
+
 // TODO: secp256k1Blake160 types
 
 export declare const secp256k1Blake160Multisig: {
@@ -15,16 +33,12 @@ export declare const secp256k1Blake160Multisig: {
    */
   transfer(
     txSkeleton: TransactionSkeleton,
-    fromAddress: string,
-    fromInfo: {
-      R: number,
-      M: number,
-      publicKeyHashes: string[],
-    },
-    toAddress: string | undefined,
-    amount: string | bigint,
+    fromAddress: Address,
+    fromInfo: MultisigScript,
+    toAddress: Address | undefined,
+    amount: bigint,
     options: {
-      config: any, // TODO: define this type later
+      config: Config,
       requiredToAddress: boolean,
     },
   ): TransactionSkeleton,
@@ -40,15 +54,11 @@ export declare const secp256k1Blake160Multisig: {
    */
   payFee(
     txSkeleton: TransactionSkeleton,
-    fromAddress: string,
-    fromInfo: {
-      R: number,
-      M: number,
-      publicKeyHashes: string[],
-    },
-    amount: string | bigint,
+    fromAddress: Address,
+    fromInfo: MultisigScript,
+    amount: bigint,
     options: {
-      config: any
+      config: Config,
     },
   ): TransactionSkeleton,
 
@@ -61,7 +71,7 @@ export declare const secp256k1Blake160Multisig: {
   prepareSigningEntries(
     txSkeleton: TransactionSkeleton,
     options: {
-      config: any,
+      config: Config,
     },
   ): TransactionSkeleton,
 
@@ -70,11 +80,7 @@ export declare const secp256k1Blake160Multisig: {
    * @param params multisig script params
    * @returns serialized multisig script
    */
-  serializeMultisigScript(params: {
-    R: number,
-    M: number,
-    publicKeyHashes: string[],
-  }): string,
+  serializeMultisigScript(params: MultisigScript): string,
 
   /**
    *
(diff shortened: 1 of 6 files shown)


packages/common-scripts/index.d.ts
```diff
@@ -0,0 +1,85 @@
...
+   * @param fromAddress must be multisig script address
+   * @param fromInfo multisig parameters
+   * @param toAddress can be any type of lock script and can left empty
+   * @param amount transfer CKB capacity in shannon
+   * @param options
+   */
+  transfer(
+    txSkeleton: TransactionSkeleton,
+    fromAddress: string,
+    fromInfo: {
```
xxuejie, line 19: I think it is better to define a separate type here for fromInfo, we can also explain what R and M does in the type

packages/common-scripts/index.d.ts
```diff
@@ -0,0 +1,85 @@
...
+      R: number,
+      M: number,
+      publicKeyHashes: string[],
+    },
+    toAddress: string | undefined,
```
xxuejie, line 24: Let's define an Address type which is just a string for now, later we can add more checking

packages/common-scripts/index.d.ts
```diff
@@ -0,0 +1,85 @@
...
+    amount: string | bigint,
```
xxuejie, line 25: For now let's start strict: we only accept BigInt here

packages/common-scripts/index.d.ts
```diff
@@ -0,0 +1,85 @@
...
+    options: {
+      config: any, // TODO: define this type later
```
xxuejie, line 27: We can define a Config type now which is just any, that will make future changes easier

packages/common-scripts/lib/secp256k1_blake160.js
```diff
@@ -196,7 +195,7 @@ async function transfer(
     }
     let witness = txSkeleton.get("witnesses").get(firstIndex);
     const newWitnessArgs = {
-      /* 65 zeros in hex */
+      /* 65 bytes zeros in hex */
```
xxuejie, line 198: `65-byte` or `65 bytes'`

packages/common-scripts/lib/secp256k1_blake160_multisig.js
```diff
@@ -0,0 +1,372 @@
...
+    template.hash_type !== script.hash_type
+  ) {
+    throw new Error(
+      "Provided script is not SECP256K1_BLAKE160_MULTISIG script!"
+    );
+  }
+}
+
+/**
+ * secp256k1_blake160_multisig requires S, R, M, N and public key hashes
```
xxuejie, line 31: For now let's put all js/ts docs in TypeScript definition file.

packages/helpers/lib/index.js
```diff
@@ -32,6 +34,8 @@ function minimalCellCapacity(fullCell, { validate = true } = {}) {
   let bytes = 8;
   bytes += new Reader(fullCell.cell_output.lock.code_hash).length();
   bytes += new Reader(fullCell.cell_output.lock.args).length();
+  // hash_type field
+  // hash_type field
```
xxuejie, line 38: Remove one line?

packages/common-scripts/index.d.ts
classicalliu: Add an interface `MultisigScript`

packages/common-scripts/index.d.ts
classicalliu: updated

packages/common-scripts/index.d.ts
classicalliu: updated

packages/common-scripts/index.d.ts
classicalliu: updated

packages/common-scripts/lib/secp256k1_blake160.js
classicalliu: updated

packages/common-scripts/lib/secp256k1_blake160_multisig.js
classicalliu: Moved to `.d.ts`

packages/helpers/lib/index.js
classicalliu: updated

classicalliu:
`secp256k1Blake160Multisig.transfer` only need `fromAddress` or  `fromInfo`
If both are filled, will check if `fromAddress` and `fromInfo` match.
If only `fromInfo`, it works well.
If only `fromAddress`, it will throw error if no inputs from this address before, for witness will need  `fromInfo`.

xxuejie:
> `secp256k1Blake160Multisig.transfer` only need `fromAddress` or `fromInfo`
> If both are filled, will check if `fromAddress` and `fromInfo` match.
> If only `fromInfo`, it works well.
> If only `fromAddress`, it will throw error if no inputs from this address before, for witness will need `fromInfo`.

I think we can define a FromInfo type like following:

```
type FromInfo = MultisigScript | Address
```

In pure JavaScript, `MultisigScript` is an object, `Address` is a string so it won't be hard to tell one from the other. This can ensure the user at least uses one of them, we can rely on typing instead of comments to regulate the API.

classicalliu:
> > `secp256k1Blake160Multisig.transfer` only need `fromAddress` or `fromInfo`
> > If both are filled, will check if `fromAddress` and `fromInfo` match.
> > If only `fromInfo`, it works well.
> > If only `fromAddress`, it will throw error if no inputs from this address before, for witness will need `fromInfo`.
> 
> I think we can define a FromInfo type like following:
> 
> ```
> type FromInfo = MultisigScript | Address
> ```
> 
> In pure JavaScript, `MultisigScript` is an object, `Address` is a string so it won't be hard to tell one from the other. This can ensure the user at least uses one of them, we can rely on typing instead of comments to regulate the API.

updated

PR #15, opened 2020-06-11
feat: Use iterator to fetch transaction one at a time

### LINKED ISSUE #3: Hanging for transactionCollector#collect to finish all the yields
I am having a test to query all the transactions with the lock args `0xdde7801c073dfb3464c7b1f05b806bb2bbb84e99` using the `transactionCollector.collect()` generator, but it seems to have to take a very long time to finish the query. I started the process 10+ mins ago, and still not finish yet. 

Meanwhile, I noticed the CKB node seems to be busy handling the requests initiated from lumos as it is utilizing 100% for one core of the CPU.

Any ideas on what is happening behind the scene?

Code: https://github.com/katat/lumos-performance-test/blob/2afa9f043e37865f97c3e042933076b1d707aae1/index.js#L41

xxuejie:
This should fix #3

### DIFF
--- a/packages/indexer/lib/index.js
+++ b/packages/indexer/lib/index.js
@@ -74,19 +74,19 @@ class Indexer {
     );
   }
 
-  getTransactionsByLockScript(script, { validateFirst = true } = {}) {
-    return this._getTransactionsByScript(script, 0, validateFirst);
+  getTransactionsByLockScriptIterator(script, { validateFirst = true } = {}) {
+    return this._getTransactionsByScriptIterator(script, 0, validateFirst);
   }
 
-  getTransactionsByTypeScript(script, { validateFirst = true } = {}) {
-    return this._getTransactionsByScript(script, 1, validateFirst);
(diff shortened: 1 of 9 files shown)


packages/indexer/lib/index.js
```diff
@@ -207,28 +207,47 @@ class TransactionCollector {
...
-        yield tx;
-      } else {
-        yield tx.transaction;
+      });
+      while (true) {
+        const hash = iter.next();
+        if (!hash) {
+          break;
+        }
+        if (hashes.has(hash)) {
```
classicalliu, line 239: Forget to add hashes ?

xxuejie: Here we don't have to insert the hashes. The set is only enabling us to skip those transactions that already have matched lock scripts. Querying only based on type scripts here, won't result in duplicate items.

classicalliu: OK, I misunderstood before.

PR #17, opened 2020-06-11
feat: Add Nervos DAO generation support

classicalliu:


packages/common-scripts/index.d.ts
```diff
@@ -1,7 +1,9 @@
-import { TransactionSkeleton } from "@ckb-lumos/types"
+import { TransactionSkeleton } from "@ckb-lumos/helpers"
 
 export type Address = string
 export type Config = any // TODO: define this type later
+export type FullCell = any
```
xxuejie, line 5: Let's take some time to define FullCell here, Config might need more thinking, hence it is a pending item, but FullCell can be defined now.

packages/common-scripts/index.d.ts
```diff
@@ -1,7 +1,9 @@
...
+export type CellProvider = any
```
xxuejie, line 6: `CellProvider` can also be defined now:

1. Let's define a `QueryOptions` structure that might take lock, type, typeIsNull and data.
2. A `CellProvider` is just one type with a `collector` method that takes a `QueryOptions`, and emits a [CellCollector](https://github.com/xxuejie/ckb-js-toolkit/blob/master/index.d.ts#L132)

packages/common-scripts/index.d.ts
classicalliu: updated

packages/common-scripts/index.d.ts
classicalliu: updated

packages/common-scripts/index.d.ts
```diff
@@ -86,4 +168,114 @@ export declare const secp256k1Blake160Multisig: {
...
+      config: Config,
+    },
+  ): Promise<TransactionSkeleton>,
+
+  /**
+   * list DAO cells,
+   *
+   * @param cellProvider
+   * @param fromAddress
+   * @param cellType "deposit" or "withdraw", can using "" to list all.
```
xxuejie, line 232: Whenever you need comment like this, please create a new type with the 2 possible values.

packages/common-scripts/lib/secp256k1_blake160.js
```diff
@@ -158,6 +160,7 @@ async function transfer(
       ) {
         continue;
       }
+
```
xxuejie, line 163: Please clean up changes like this.

packages/types/index.d.ts
```diff
@@ -1 +1,63 @@
...
+    type?: Script;
+  };
+  data: string;
+  out_point: OutPoint;
+  block_hash: string;
+  block_number: string;
+}
+
+export interface QueryOptions {
+  lock?: Script | null;
```
xxuejie, line 32: I think we can just allow `Script` or `undefined`, and reject `null`

packages/common-scripts/index.d.ts
classicalliu: OK, updated now.

packages/common-scripts/lib/secp256k1_blake160.js
classicalliu: updated.

packages/types/index.d.ts
classicalliu: updated

PR #24, opened 2020-06-17
feat: Config manager implementation

xxuejie:


packages/config-manager/lib/index.js
```diff
@@ -0,0 +1,86 @@
...
+    throw new Error(`${debugPath} must be a hex string of 66 bytes long!`);
+  }
+}
+
+function assertInteger(debugPath, i) {
+  if (i === "0x0") {
+    return;
+  }
+  if (!/^0x[1-9a-fA-F][0-9a-fA-F]*$/.test(i)) {
+    throw new Error(`${debugPath} must be a hex integer!`);
```
classicalliu, line 24: Is `0x1` a valid hex integer ? It seems must be `0x01` ?

xxuejie: We must use `0x1` here to represent a number, no leading zeros are allowed.

PR #32, opened 2020-06-24
feat: SQL backend indexer

xxuejie:
This PR implements a different indexer that uses SQL as a backend. For simplicity, it is implemented as pure JS right now, later we might rewrite the core in Rust for better performance.

### DIFF
--- a/packages/sql-indexer/package.json
+++ b/packages/sql-indexer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ckb-lumos/sql-indexer",
-  "version": "0.3.1",
+  "version": "0.3.2",
   "description": "SQL based CKB Indexer",
   "author": "Xuejie Xiao <xxuejie@gmail.com>",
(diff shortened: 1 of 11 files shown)


packages/sql-indexer/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@ckb-lumos/sql-indexer",
+  "version": "0.3.1",
+  "description": "SQL based CKB Indexer using typeorm",
```
classicalliu:
```suggestion
  "description": "SQL based CKB Indexer using knex",
```

classicalliu (approved):

packages/sql-indexer/package.json
xxuejie: Remove the whole `using` part since we might rewrite it later without knex

classicalliu (approved):

PR #45, opened 2020-07-02
feat: Add block range parameter to transaction iterator

bitrocks:
This PR is for https://github.com/nervosnetwork/lumos/issues/4 to support `#getTransactionsByLockScript(fromBlock, toBlock)`


packages/indexer/lib/index.js
```diff
@@ -191,7 +193,7 @@ class TransactionCollector {
   constructor(
     indexer,
     { lock = null, type = null } = {},
-    { skipMissing = false, includeStatus = true } = {}
+    { skipMissing = false, includeStatus = true, fromBlock = null, toBlock = null } = {},
```
xxuejie: `fromBlock` and `toBlock` should be in the query object together with lock and type, not options here.

xxuejie (commented):
You need to bump the version of indexer package for the CI to pass, since native code is modified here.

packages/indexer/lib/index.js
bitrocks: ok, I have move the `fromBlock | toBlock` to the same field in `lock | type`.

packages/indexer/lib/index.js
```diff
@@ -190,8 +192,8 @@ class CellCollector {
 class TransactionCollector {
   constructor(
     indexer,
-    { lock = null, type = null } = {},
-    { skipMissing = false, includeStatus = true } = {}
+    { lock = null, type = null, fromBlock = null, toBlock = null } = {},
```
xxuejie, line 195: Please update `index.d.ts` file with the new available parameters. Make sure to document that toBlock is exclusive as well.

bitrocks: The `index.d.ts` has been updated.  Besides, I also changed the `toBlock` related logic to  make it  included in the query range, which might cause less confusion for users to read the docs and call the API.

packages/base/index.d.ts
```diff
@@ -80,6 +80,8 @@ export interface QueryOptions {
   type?: Script | "empty";
   data?: string;
   argsLen?: number;
+  fromBlock?: number;
+  toBlock?: number;
```
xxuejie, line 85: Add a typedoc comment denoting this is inclusive. I know you have included a comment in the README part, but here we were generating API documents powered by typedoc, it's worth mentioning it here.

packages/indexer/README.md
```diff
@@ -5,6 +5,7 @@ CKB indexer used in lumos framework. Might be possible for independent usage as
 The indexer is designed to consume from the following sources:
 
 * Direct access of CKB's data dir via RocksDB's readonly or secondary mode;
+
```
xxuejie: Remove this.

packages/indexer/native/src/lib.rs
```diff
@@ -435,11 +437,25 @@ declare_types! {
...
+                end_key.extend_from_slice(&to_block_number.to_be_bytes());
+            } else {
+                end_key.extend_from_slice(&u64::MAX.to_be_bytes());
+            }
             let iter = store.iter(&start_key, IteratorDirection::Forward);
             if iter.is_err() {
                 return cx.throw_error("Error creating iterator!");
             }
-            let iter = iter.unwrap().take_while(move |(key, _)| key.starts_with(&start_key));
+            let iter = iter.unwrap().take_while(move |(key, _)| key.to_vec() < end_key);
```
xxuejie, line 458: What about using `<=` here, so we won't need to add 1 to `to_block_number` above? Although the possibility is extremely rare, adding 1 without bound check is dangerous, it also will generate warnings when we use certain automatic code analysis tools. I doubt this will be a real issue, it is more of a cosmetic fix IMO.

packages/indexer/package.json
```diff
@@ -56,4 +56,4 @@
   "devDependencies": {
     "electron-build-env": "^0.2.0"
   }
-}
+}
```
xxuejie: Please fix this.

packages/indexer/native/src/lib.rs
bitrocks: I considered using `<=` in the first place, but decided not after checking the `TxLockScript | TxTypeScript` serialization here, https://github.com/quake/ckb-indexer/blob/94b4db3a5e793999d970c838e3843c72ef164e8a/src/indexer.rs#L97. 

There might be a inconsistent situation: 
1. bytes after `block_number` are all-zeros, which will be included in the result;
2. bytes after `block_number` are not all-zeros, which will not;

Keys with the same prefix `script + to_block_number` but be divided into two spaces with `<=` conditions, could cause unexpected query result.

xxuejie: Ah I see, yes you are right here. In that sense ignore this.

xxuejie (approved):

packages/base/index.d.ts
bitrocks: Updated.

packages/indexer/README.md
```diff
@@ -58,6 +58,27 @@ for await (const tx of txCollector.collect()) {
...
+txCollector = new TransactionCollector(indexer, {
+  lock: {
+    code_hash: 
+      "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
+    hash_type: "type",
+    args: "0xa528f2b9a51118b193178db4cf2f3db92e7df323",
+  },
+  fromBlock: 1000,
+  toBlock: 10000,
+});
```
bitrocks, line 73: As the lock script showed in original doc is only appeared in genesis block so far, which is too trivial for a range query demo, so here replace it with cell base output lock script which  appeared in over 290k transactions.

Issue #16, opened 2020-06-11
Further split lock and type in TransactionCollector


xxuejie:
Right now TransactionCollector uses the same queries `lock` and `type` as CellCollector. While they are enough for cells, the granularity would be too big for transactions. As a result, we would want to split `lock` and `type` into the following queries:

* `input_lock`: transactions whose input cells use specified lock
* `output_lock`: transactions whose output cells use specified lock
* `input_type`: transactions whose input cells use specified type
* `output_type`: transactions whose output cells use specified type.

PR #46, opened 2020-07-03
feat: Split script with io_type in transaction collector

bitrocks:
This PR is for #16

### DIFF
--- a/packages/base/index.d.ts
+++ b/packages/base/index.d.ts
@@ -30,11 +30,6 @@ export interface Script {
   args: HexString;
 }
 
-export interface ScriptWrapper {
-  script: Script;
-  io_type: "input" | "output" | "both";
-}
-
 export interface OutPoint {
   tx_hash: Hash;
   index: HexString;
@@ -88,6 +83,11 @@ export interface QueryOptions {
   fromBlock?: number;
   /** `toBlock` itself is included in range query. */
   toBlock?: number;
+}
+
+export interface ScriptWrapper {
+  script: Script;
+  ioType: "input" | "output" | "both";
 }
 
 export interface CellCollectorResults {
--- a/packages/base/index.d.ts
+++ b/packages/base/index.d.ts
@@ -32,7 +32,7 @@ export interface Script {
 
 export interface ScriptWrapper {
   script: Script;
-  io_type: "input" | "output";
+  io_type: "input" | "output" | "both";
 }
 
 export interface OutPoint {
--- a/packages/base/index.d.ts
+++ b/packages/base/index.d.ts
@@ -30,6 +30,11 @@ export interface Script {
   args: HexString;
 }
 
+export interface ScriptWrapper {
+  script: Script;
+  io_type: "input" | "output";
+}
+
 export interface OutPoint {
   tx_hash: Hash;
   index: HexString;
@@ -75,8 +80,8 @@ export interface Cell {
   block_number: HexString;
 }
 export interface QueryOptions {
-  lock?: Script;
-  type?: Script | "empty";
+  lock?: Script | ScriptWrapper;
+  type?: Script | ScriptWrapper | "empty";
   data?: string;
   argsLen?: number;
   /** `fromBlock` itself is included in range query. */
--- a/packages/indexer/README.md
+++ b/packages/indexer/README.md
@@ -58,7 +58,7 @@ for await (const tx of txCollector.collect()) {
 }
 ```
 
-Query transactions between given `block_number` range is supported:
+Range query for transactions between given block_numbers is supported:
 
 ```javascript
 txCollector = new TransactionCollector(indexer, {
@@ -68,8 +68,8 @@ txCollector = new TransactionCollector(indexer, {
     hash_type: "type",
     args: "0xa528f2b9a51118b193178db4cf2f3db92e7df323",
   },
-  fromBlock: 1000,
-  toBlock: 10000,
+  fromBlock: 0,
+  toBlock: 2000,
 });
 
 for await (const tx of txCollector.collect()) {
@@ -78,6 +78,30 @@ for await (const tx of txCollector.collect()) {
 ```
 
 It will fetch transactions between `[fromBlock, toBlock]`, which means both `fromBlock` and `toBlock` are included in query range.
+
+Fine grained query for transactions by scripts with `io_type` is supported: 
+
+```javascript
+txCollector = new TransactionCollector(indexer, {
+  lock: {
+    script: {
+      code_hash: 
+        "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
+      hash_type: "type",
+      args: "0xa528f2b9a51118b193178db4cf2f3db92e7df323",
+    },
+    io_type: "input",
+  },
+  fromBlock: 0,
+  toBlock: 2000,
+});
+
+for await (const tx of txCollector.collect()) {
+  console.log(tx);
+}
+```
+
+The `io_type` field is among `input | output | both`.
 
 ## Electron note
 
--- a/packages/indexer/README.md
+++ b/packages/indexer/README.md
@@ -79,7 +79,7 @@ for await (const tx of txCollector.collect()) {
 
 It will fetch transactions between `[fromBlock, toBlock]`, which means both `fromBlock` and `toBlock` are included in query range.
 
-Fine grained query for transactions by scripts with `io_type` is supported: 
+Fine grained query for transactions by scripts with `ioType` is supported: 
 
 ```javascript
 txCollector = new TransactionCollector(indexer, {
@@ -90,7 +90,7 @@ txCollector = new TransactionCollector(indexer, {
       hash_type: "type",
       args: "0xa528f2b9a51118b193178db4cf2f3db92e7df323",
     },
-    io_type: "input",
+    ioType: "input",
   },
   fromBlock: 0,
   toBlock: 2000,
@@ -101,7 +101,7 @@ for await (const tx of txCollector.collect()) {
 }
 ```
 
-The `io_type` field is among `input | output | both`.
+The `ioType` field is among `input | output | both`.
 
 ## Electron note
 
--- a/packages/indexer/lib/index.js
+++ b/packages/indexer/lib/index.js
@@ -52,6 +52,7 @@ class Indexer {
     return this.nativeIndexer.getTransactionsByScriptIterator(
       normalizers.NormalizeScript(script),
       scriptType,
+      ioType,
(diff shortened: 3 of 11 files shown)


packages/indexer/lib/index.js
```diff
@@ -190,88 +191,90 @@ class CellCollector {
 class TransactionCollector {
   constructor(
     indexer,
-    { lock = null, type = null } = {},
+    { input_lock = null, output_lock = null, input_type = null, output_type } = {},
```
xxuejie: We need to also keep the original `lock` and `type` query, for compatibility reasons.

bitrocks: In stead of keepping both `lock | type | input_lock | output_lock | input_type | output_type` to enable a fine-grained query, I would propose to add new field in existing `lock | type` : 

```
lock: {
  script: {
    code_hash:
        "0x0000000000000000000000000000000000000000000000000000000000000000",
      hash_type: "data",
      args: "0x62e907b15cbf27d5425399ebf6f0fb50ebb88f18",
  },
 io_type: null // or 0 - input, 1 - output
}
```

It seems like a more natural way to support #16

xxuejie: 2 things:

1. We cannot break compatibility here, so if you do this path, lock should accept both a plain script, or an object containing script and `io_type` like shown above.
2. `io_type` is an internal detail, we shouldn't expose the values out here, it's better to use `"input"` and `"output"`, which will have better typing support at TypeScript side.

xxuejie:
@bitrocks Please resolve the conflicts here

packages/indexer/lib/index.js
```diff
@@ -198,12 +205,23 @@ class TransactionCollector {
     if (!lock && !type) {
       throw new Error("Either lock or type script must be provided!");
     }
-    if (lock) {
+
+    if (lock && !lock.script) {
```
xxuejie, line 202: There is a lot of logic duplication, I would suggest refactoring the logic for telling Script from ScriptWrapper in a separate place, so the constructor can deal with this change. After that, the code below should not need to test for the existence of `script` field again and again.

One idea would be adding a `both` type to `io_type`, so we can convert all Script to ScriptWrapper.

packages/base/index.d.ts
```diff
@@ -30,6 +30,11 @@ export interface Script {
   args: HexString;
 }
 
+export interface ScriptWrapper {
+  script: Script;
+  io_type: "input" | "output";
```
xxuejie: We should use camel case here following node's convention.

packages/indexer/native/src/lib.rs
```diff
@@ -430,32 +432,44 @@ declare_types! {
...
+            let script_type = cx.argument::<JsNumber>(2)?.value() as u32;
+            let prefix = if script_type == 1 {
                 KeyPrefix::TxTypeScript
             } else {
                 KeyPrefix::TxLockScript
             };
             let mut start_key = vec![prefix as u8];
             start_key.extend_from_slice(&script.as_slice()[SCRIPT_SERIALIZE_OFFSET..]);
+
+            let io_type= cx.argument::<JsValue>(3)?;
```
xxuejie: `io_type = ...`

packages/indexer/native/src/lib.rs
```diff
@@ -430,32 +432,44 @@ declare_types! {
                 return cx.throw_error(format!("Error assembling script: {:?}", script.unwrap_err()));
             }
             let script = script.unwrap();
-            let prefix = if cx.argument::<JsNumber>(2)?.value() as u32 == 1 {
+            let script_type = cx.argument::<JsNumber>(2)?.value() as u32;
```
xxuejie: I don't see `script_type` getting used below, why creating a variable?

packages/indexer/native/src/lib.rs
```diff
@@ -430,32 +432,44 @@ declare_types! {
...
-            let to_block = cx.argument::<JsValue>(4)?;
+            let to_block = cx.argument::<JsValue>(5)?;
             if to_block.is_a::<JsNumber>() {
                 // here set to_block_number as toBlock + 1, making the toBlock included in query range.
                 let to_block_number = to_block.downcast::<JsNumber>().or_throw(&mut cx)?.value() as u64 + 1;
                 end_key.extend_from_slice(&to_block_number.to_be_bytes());
             } else {
                 end_key.extend_from_slice(&u64::MAX.to_be_bytes());
             }
+
```
xxuejie, line 458: Remove this.

packages/indexer/native/src/lib.rs
```diff
@@ -430,32 +432,44 @@ declare_types! {
...
             let iter = store.iter(&start_key, IteratorDirection::Forward);
             if iter.is_err() {
                 return cx.throw_error("Error creating iterator!");
             }
-            let iter = iter.unwrap().take_while(move |(key, _)| key.to_vec() < end_key);
+            // FIXME
```
xxuejie: What is this?

packages/base/index.d.ts
bitrocks: But it seems all property fields in the `index.d.ts`  is snake case? Am I misunderstanding it?

packages/indexer/lib/index.js
bitrocks: I have refactored this part and converted all Script to ScriptWrapper in constructor. Besides, I extract a `get_transaction_hashes` method for the shared logic of fetching all transaction_hashes from transaction iterator.

packages/indexer/native/src/lib.rs
bitrocks: It's used one line below the variable declaration, for deciding the `prefix`.

packages/base/index.d.ts
xxuejie: Only those fields that are directly coming from CKB RPC API are in snake case, this includes most CKB data structures. For APIs that are specific to lumos, we will use camel case.

packages/indexer/native/src/lib.rs
bitrocks: There was a bug when I push the code and it's been solved in the latest commit.

packages/indexer/native/src/lib.rs
bitrocks: updated.

packages/indexer/native/src/lib.rs
bitrocks: This empty line is added here(and maybe somewhere else) for ergonomics reasons when I am looking at the long and crowded method body.

packages/base/index.d.ts
```diff
@@ -30,6 +30,11 @@ export interface Script {
   args: HexString;
 }
 
+export interface ScriptWrapper {
+  script: Script;
+  io_type: "input" | "output" | "both";
```
xxuejie: This is still `io_type`

packages/indexer/native/src/lib.rs
xxuejie: Yes but this is the only place we are using `script_type` here, I was curious why embedding it in `prefix` calculation won't work? What's the rationale behind splitting it to a varaible?

bitrocks: Get it. I'll merge them into a single line in the next commit.

packages/base/index.d.ts
bitrocks: Will be fixed in the next commit.

packages/indexer/README.md
```diff
@@ -79,6 +79,30 @@ for await (const tx of txCollector.collect()) {
...
+```javascript
+txCollector = new TransactionCollector(indexer, {
+  lock: {
+    script: {
+      code_hash: 
+        "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
+      hash_type: "type",
+      args: "0xa528f2b9a51118b193178db4cf2f3db92e7df323",
+    },
+    io_type: "input",
```
xxuejie: Here we are still using snake case.

packages/indexer/lib/index.js
```diff
@@ -198,116 +198,92 @@ class TransactionCollector {
     if (!lock && !type) {
       throw new Error("Either lock or type script must be provided!");
     }
-    if (lock) {
+    // Wrap the plain `Script` to `ScriptWrapper`.
+    if (lock && !lock.script) {
       validators.ValidateScript(lock);
+      this.lock = { script: lock, io_type: "both" };
+    }
+    if (lock && lock.script) {
```
xxuejie: We can either execute line 202 - 205, or line 206 - 209, here we are doing script validation twice.

packages/indexer/lib/index.js
```diff
@@ -198,116 +198,92 @@ class TransactionCollector {
...
+
+    if (this.lock) {
+      let script_type = 0;
+      lockHashes = new OrderedSet(
         this.indexer
           ._getTransactionsByScriptIterator(
-            this.lock,
-            0,
+            this.lock.script,
+            script_type,
```
xxuejie, line 237: Why give this a variable if you will never modify/use it?

packages/indexer/lib/index.js
```diff
@@ -198,116 +198,92 @@ class TransactionCollector {
...
             this.fromBlock,
             this.toBlock
           )
-          .collect()
+          .collect(this.lock.io_type)
```
xxuejie: This is still snake case.

packages/indexer/README.md
bitrocks: Sorry for missing these `io_type`. Will fix it.

packages/indexer/lib/index.js
bitrocks: I think give it an identifier is more readable than raw `0|1` , and yes, it should be a const type here. Will fix it.

packages/indexer/lib/index.js
bitrocks: You mean here doing condition check twice? The `ValidateScript` part would be executed only once.  I would use `else if` to concat those two seperate part.

xxuejie: Think about this:

1. We provide a plain script here.
2. Line 202 - 205 will validate the script, then make it a script wrapper, and save it to `this.lock`
3. At line 206, the code finds that `this.lock` is a script wrapper, then enters the if
4. Line 207 validates the script in the script wrapper again.

bitrocks: I don't think it will pass the second condition check if `this.lock` is set in line 202 - 205,  the `lock` in 206 is different thing from `this.lock` ?

xxuejie: Oh yes, you are right, my mistake here.

xxuejie (approved):

PR #47, opened 2020-07-06
Migrate helpers to TypeScript

classicalliu:


.eslintrc.js
```diff
@@ -8,7 +15,7 @@ module.exports = {
     "SharedArrayBuffer": "readonly"
   },
   "rules": {
-    "no-unused-vars": "error",
-    "no-var": "error"
+    "no-var": "error",
+    "@typescript-eslint/no-unused-vars": ["error"]
```
xxuejie, line 19: I think we will also need to preserve the original `no-unused-vars` rule for JS code.

packages/base/index.d.ts
```diff
@@ -49,12 +51,18 @@ export interface Input {
 export interface Output {
   capacity: HexString;
   lock: Script;
-  type?: Script;
+  type?: Script | null;
```
xxuejie: I thin we should either use `?:` syntax, or `| null`, but we should not use both.

packages/helpers/package.json
```diff
@@ -6,26 +6,29 @@
   "homepage": "https://github.com/nervosnetwork/lumos#readme",
   "license": "MIT",
   "main": "lib/index.js",
-  "types": "index.d.ts",
+  "types": "lib/index.d.ts",
```
xxuejie, line 9: If this is a TypeScript package, maybe this part is not necessary?

packages/helpers/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
```
xxuejie, line 1: Let's keep a single `tsconfig.json` in root folder for all packages.

.eslintrc.js
classicalliu: This rule also works with `js` files.

If add origin `no-unused-vars` back, will get two similar errors in `js` files.

```
  2:17  error  'validators' is assigned a value but never used  no-unused-vars
  2:17  error  'validators' is assigned a value but never used  @typescript-eslint/no-unused-vars
```

and in `ts` files

```
  16:29  error  'Set' is defined but never used  no-unused-vars
  16:29  error  'Set' is defined but never used  @typescript-eslint/no-unused-vars
```

And if I only use original `no-unused-vars`, some imported variables only used for types will cause this error.

```
  1:10  error  'Address' is defined but never used  no-unused-vars
  1:19  error  'Script' is defined but never used   no-unused-vars
```

packages/base/index.d.ts
classicalliu: Is this mean we can't both use `undefined` and `null` to express empty?

If so, I will preserve `null`, and remove `?:`, for `rpc` returns `null`, but it should explicit declaration this key is `null`.

.eslintrc.js
xxuejie: Gotcha.

packages/base/index.d.ts
xxuejie: I think we should stick to one across the entire lumos codebase, I don't mind whether it is `null` or `undefined`, but I do mind that we use only one.

packages/helpers/package.json
classicalliu: This package can reference from `js` and `ts`, and source typescript files will compile to `*.js` and `*.d.ts` files when publish, and exports them. It still use `js` and `.d.ts` pairs.

packages/helpers/tsconfig.json
classicalliu: I'll try to move to root folder.

packages/base/index.d.ts
classicalliu: ok

classicalliu: `null` removed.

packages/helpers/tsconfig.json
classicalliu: Add a global `tsconfig` in root folder, and extends global `tsconfig` in `helpers`.

I don't find the way to solve the `outDir` problem, so a add a `tsconfig` and extends global one.

xxuejie (approved):
Let's add a prepublish task to invoke build in helpers. Then this is all good.

classicalliu:
> Let's add a prepublish task to invoke build in helpers. Then this is all good.

```
npm WARN prepublish-on-install As of npm@5, `prepublish` scripts are deprecated.
npm WARN prepublish-on-install Use `prepare` for build steps and `prepublishOnly` for upload-only.
npm WARN prepublish-on-install See the deprecation note in `npm help scripts` for more information.
```

Since `prepublish` is deprecated, I'm add a `prepublishOnly` task for `helpers` to invode `clean` and `build` before `npm publish`.

xxuejie (approved):

PR #49, opened 2020-07-07
feat: Add more query options(skip, fromBlock, toBlock) to CellCollector

bitrocks:
This PR is for #3 to support page jump.

So far, I only add the `skip` option in `TransactionCollector`, if the implementation seems reasonable for you, I would complete the `CellCollector` part then.  @xxuejie

### DIFF
--- a/packages/indexer/native/src/lib.rs
+++ b/packages/indexer/native/src/lib.rs
@@ -365,41 +365,37 @@ declare_types! {
             } else {
                 KeyPrefix::CellLockScript
             };
-
             let args_len = cx.argument::<JsNumber>(3)?.value();
             if args_len > u32::max_value() as f64 {
                 return cx.throw_error("args length must fit in u32 value!");
             }
-            let from_block = cx.argument::<JsValue>(4)?;
-            let to_block = cx.argument::<JsValue>(5)?;
             let args_len = if args_len <= 0.0 {
                 script.args().len() as u32
             } else {
-                if from_block.is_a::<JsNumber>() || to_block.is_a::<JsNumber>() {
-                    return cx.throw_error("prefix search on args and range query can't be used at the same time!");
-                }
+                // when prefix search on args, the args parameter is be shorter than actual args, so need set the args_len manully.
                 args_len as u32
             };
-
             let mut start_key = vec![prefix as u8];
             start_key.extend_from_slice(script.code_hash().as_slice());
             start_key.extend_from_slice(script.hash_type().as_slice());
             start_key.extend_from_slice(&args_len.to_le_bytes());
             start_key.extend_from_slice(&script.args().raw_data());
 
-            let mut end_key = start_key.clone();
-
-            if from_block.is_a::<JsNumber>() {
+            let from_block = cx.argument::<JsValue>(4)?;
+            let to_block = cx.argument::<JsValue>(5)?;
+            let from_block_number_slice = if from_block.is_a::<JsNumber>() {
                 let from_block_number = from_block.downcast::<JsNumber>().or_throw(&mut cx)?.value() as u64;
-                start_key.extend_from_slice(&from_block_number.to_be_bytes());
-            }
-            if to_block.is_a::<JsNumber>() {
+                from_block_number.to_be_bytes()
+            } else {
+                0_u64.to_be_bytes()
+            };
+            let to_block_number_slice = if to_block.is_a::<JsNumber>() {
                 // here set to_block_number as toBlock + 1, making the toBlock included in query range.
                 let to_block_number = to_block.downcast::<JsNumber>().or_throw(&mut cx)?.value() as u64 + 1;
-                end_key.extend_from_slice(&to_block_number.to_be_bytes());
+                to_block_number.to_be_bytes()
             } else {
-                end_key.extend_from_slice(&u64::MAX.to_be_bytes());
-            }
+                u64::MAX.to_be_bytes()
+            };
 
             let skip = cx.argument::<JsValue>(6)?;
             let skip_number = if skip.is_a::<JsNumber>() {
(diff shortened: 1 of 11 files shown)


packages/indexer/native/src/lib.rs
```diff
@@ -488,14 +488,25 @@ declare_types! {
...
+            } else {
+                0_u64
+            };
+
             let mut this = cx.this();
             let hashes = {
                 let guard = cx.lock();
                 let mut iterator = this.borrow_mut(&guard);
                 let mut hashes = vec![];
+                let mut current_counter = 0;
```
xxuejie: Rust's iterator has a [skip](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip) method, can we leverage that? I think it's better here we first use [filter](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter) to grab all keys that ends with `io_type_mark`, then use skip to locate the correct offset.

xxuejie (commented):
The general workflow looks fine, just one comment on actual implementation

packages/indexer/native/src/lib.rs
bitrocks: The pattern sounds feasible and more idiomatic. I'll give it a try.

bitrocks:
I have add `skip` to both `TransactionCollector` and `CellCollector`, and add `fromBlock | toBlock` to `CellCollector`, and update the related docs. But there are some problems need to be solved:

1. It seems infeasible for prefix search on args and range query in `CellCollector` to work at the same time,  cause the `start_key` would have an incontinuous part in the middle.  Maybe it can be solved by construct a more complex `predicate` in iterator, but I haven't try it so far.

2. When refactor the `CellCollector`,  I've encountered some confusions:  what's the purpose of wrapping the `outPoint` in `BufferValue`;

xxuejie:
1. I think we can just extract from and to block number from each key, and perform the test
2. BufferValue provide equality test based on OutPoint contents

packages/indexer/native/src/lib.rs
```diff
@@ -398,6 +325,146 @@ declare_types! {
...
+            if iter.is_err() {
+                return cx.throw_error("Error creating iterator!");
+            }
+
+            let iter = iter.unwrap()
+            .take_while(move |(key, _)| {
+                let block_number_slice = key[key.len() - 16..key.len() - 8].try_into();
+                key.starts_with(&start_key) && to_block_number_slice > block_number_slice.unwrap()
+            })
+            .filter( move |(key, _)| {
```
xxuejie, line 417: There might be a performance quirk: it really depends on the implementation, but if there are like 2 million blocks, and the toBlock ends at `100`, it might be the case that the code here also needs to loop and filter against all the rest 1999900 blocks, since the iterator might not know that the keys are sorted. I think it's better we manually do an early stop.

packages/indexer/native/src/lib.rs
```diff
@@ -456,11 +532,18 @@ declare_types! {
...
+            } else {
+                0_usize
+            };
+
             let iter = store.iter(&start_key, IteratorDirection::Forward);
             if iter.is_err() {
                 return cx.throw_error("Error creating iterator!");
             }
-            let iter = iter.unwrap().take_while(move |(key, _)| key.to_vec() < end_key);
+            let iter = iter.unwrap().take_while(move |(key, _)| key.to_vec() < end_key).filter(move |(key, _)| key.ends_with(&io_type_mark)).skip(skip_number);
```
xxuejie, line 546: Same performance quirk with the above.

xxuejie (commented):
Notice the version we release after this PR should be 0.5.0.

packages/indexer/native/src/lib.rs
bitrocks: I considered to perform an early stop on `fromBlock` like last implementation. But the only possible way I came out is using another rockdb iterator with ` IteratorDirection::Reverse`, so what do you mean an early stop manually here?

packages/indexer/native/src/lib.rs
```diff
@@ -398,6 +325,146 @@ declare_types! {
...
+                let guard = cx.lock();
+                let mut iterator = this.borrow_mut(&guard);
+                let mut key_values = vec![];
+                while let Some((key, value)) = iterator.0.next() {
+                    key_values.push((key,value));
+                }
+                key_values
+            };
+            let js_out_points = JsArray::new(&mut cx, live_cell_key_values.len() as u32);
+            for (i, (key, value)) in live_cell_key_values.iter().enumerate() {
```
xxuejie, line 438: You can perform the early stop checking at here.

packages/indexer/native/src/lib.rs
bitrocks: It's slightly different from the `JsLiveCellIterator` case, cause the `io_type` is not sequencially store in rocksdb, so the one by one comparison seems unavoidable here.

xxuejie: You are right, ignore this comment here.

packages/indexer/native/src/lib.rs
bitrocks: I don't get it. The rocksdb's  `key space traveling` phase  would finish after line 432 - 434, I don't see what we can do here. Could you elaborate more on it?

A partial performance improvement I can do is separating the `JsLiveCellIterator` construct into two cases:
1. with `prefix search on args`, use the current version of iterator;
2. without `prefix search on args`, use last version of iterator, which doesn't suffer from the performance issue.

packages/indexer/native/src/lib.rs
```diff
@@ -398,6 +325,146 @@ declare_types! {
...
+            Ok(LiveCellIterator(Box::new(iter)))
+        }
+
+        method collect(mut cx) {
+            let mut this = cx.this();
+            let live_cell_key_values = {
+                let guard = cx.lock();
+                let mut iterator = this.borrow_mut(&guard);
+                let mut key_values = vec![];
+                while let Some((key, value)) = iterator.0.next() {
```
xxuejie, line 432: Ooops, I mean you can move the toBlock detection here, when you have detected a block number that is after toBlock, you can terminate the while loop here for an early stop.

packages/indexer/native/src/lib.rs
xxuejie: Wrong location, see comments above.

xxuejie (approved):
Based on offline chat, we decided to leave the performance optimization to a later PR.

PR #65, opened 2020-07-16
Sudt support

classicalliu:


packages/common-scripts/src/common.ts
```diff
@@ -86,27 +88,27 @@ export async function transfer(
   if (useLocktimeCellsFirst) {
     // if provider tipHeader
     if (tipHeader) {
-      [txSkeleton, deductAmount] = await lockTimePool.transfer(
+      const result = await lockTimePool.transfer(
```
xxuejie: Any reason for this change?

packages/common-scripts/src/sudt.ts
```diff
@@ -0,0 +1,421 @@
...
+  parseAddress,
+  minimalCellCapacity,
+  TransactionSkeletonType,
+  Options,
+} from "@ckb-lumos/helpers";
+import { Set, List } from "immutable";
+import { getConfig, Config } from "@ckb-lumos/config-manager";
+import { collectCells, LocktimeCell } from "./locktime_pool";
+
+export async function createToken(
```
xxuejie: issueToken is probably a better name.

packages/common-scripts/src/sudt.ts
```diff
@@ -0,0 +1,421 @@
...
+ * @param toAddress
+ * @param amount
+ * @param changeAddress if not provided, will use first fromInfo
+ * @param capacity
+ * @param options
+ */
+export async function transfer(
+  txSkeleton: TransactionSkeletonType,
+  fromInfos: FromInfo[],
+  sudtToken: Hash,
```
xxuejie: Let's define a `Token` type, for now it could be just the args, but later we can expand this to full Script

xxuejie (commented):
Let's also add a short introduction on how to use it in the README part.

packages/common-scripts/src/common.ts
classicalliu: I forget to revert this after debug, I'll revert changes in `transfer`, it's no need to change now.

packages/common-scripts/src/sudt.ts
classicalliu: updated.

packages/common-scripts/src/sudt.ts
classicalliu: updated.

xxuejie (approved):
Let's add a new utility method named `ownerForSudt`: it takes ` FromInfo`, and returns a `Token`. It calculates the SUDT token assuming the FromInfo is used as the owner. After this new method is added we can merge the PR.

classicalliu:
> Let's add a new utility method named `ownerForSudt`: it takes ` FromInfo`, and returns a `Token`. It calculates the SUDT token assuming the FromInfo is used as the owner. After this new method is added we can merge the PR.

Added.

xxuejie (approved):

PR #69, opened 2020-07-23
Support anyone-can-pay script

classicalliu:
Now anyone-can-pay script only support to transfer anyone-can-pay addresses to anyone-can-pay address.
And transfer to anyone-can-pay address from any lock script will integrate to `common` later.


packages/common-scripts/src/anyone_can_pay.ts
```diff
@@ -0,0 +1,549 @@
...
+import { Set, List } from "immutable";
+const { CKBHasher, ckbHash, readBigUInt128LE } = utils;
+
+export class CellCollector {
+  private cellCollector: CellCollectorType;
+  private config: Config;
+  public readonly fromScript: Script;
+
+  constructor(
+    fromAddress: Address,
```
xxuejie: Can we use `FromInfo` here?

packages/common-scripts/src/anyone_can_pay.ts
```diff
@@ -0,0 +1,549 @@
...
+    cellProvider: CellProvider,
+    {
```
xxuejie, line 43: Can we skip QueryOptions here? Is there any chance that we have to let users customize the query used here?

packages/common-scripts/src/anyone_can_pay.ts
```diff
@@ -0,0 +1,549 @@
...
+
+    txSkeleton = await setupInputCell(txSkeleton, firstInjectInputIndex, {
+      config,
+    });
+  }
+
+  return txSkeleton;
+}
+
+export async function transfer(
```
xxuejie: I feel like maybe we should not provide transfer here. I do want to distinguish transfer from withdraw in lumos:

* transfer is an action that you exact CKBs from a bunch of cells that are alike, it does not matter which cell is in fact used.
* withdraw is an action that you exact CKBs from exactly one cell, and you can only use that single cell.

I feel like in ACP, it's more likely we are dealing with withdrawing operation, not transfer operation.

packages/common-scripts/src/anyone_can_pay.ts
```diff
@@ -0,0 +1,549 @@
...
+        message: hasher.digestHex(),
+      };
+      signingEntries = signingEntries.push(signingEntry);
+    }
+  }
+  txSkeleton = txSkeleton.set("signingEntries", signingEntries);
+  return txSkeleton;
+}
+
+export async function destroyAnyoneCanPayInput(
```
xxuejie: One we change transfer to withdraw, I feel like we can merge this function with withdraw: the cell will be destroyed once one withdraw all the amount in the cell.

packages/common-scripts/src/anyone_can_pay.ts
classicalliu: OK.

packages/common-scripts/src/anyone_can_pay.ts
classicalliu: It's for `anyone-can-pay` with `sudt` https://github.com/nervosnetwork/lumos/pull/69/files#diff-2cf727f7322effda3e47a60c2bbcc253R199

If we skipped `QueryOptions`, we should provide a `sudtToken` param, for search `anyone-can-pay` lock script with `sudt` type script.

packages/common-scripts/src/anyone_can_pay.ts
classicalliu: It's this means we can withdraw a cell with withdraw capacity range `[0,cell.capcaity-minimalCellCapcity] ∪ cell.capcaity`, and withdraw amount (from data) range `[0, cell.amount]`

packages/common-scripts/src/anyone_can_pay.ts
xxuejie: Okay in that sense let's keep QueryOptions here.

packages/common-scripts/src/anyone_can_pay.ts
xxuejie: Yeah I think that's the plan.

classicalliu: ok

packages/common-scripts/src/anyone_can_pay.ts
classicalliu: Updated.

packages/common-scripts/src/anyone_can_pay.ts
classicalliu: Updated.

packages/common-scripts/src/anyone_can_pay.ts
classicalliu: Updated.

xxuejie (approved):

PR #70, opened 2020-07-25
feat: Notification mechanism support

bitrocks:
It's for https://github.com/nervosnetwork/lumos/issues/55 and previously discussed in #4.

- [x] RocksDB Indexer
- [x] SQL Indexer

The RockDB indexer's notification mechanism implementation is based on the neon's [eventhandler](https://github.com/neon-bindings/rfcs/blob/master/text/0025-event-handler.md) feature.

It will emit a `changed` event when a block containing the subsribed script is indexed or rollbacked. And developers are supposed to pull from the data sources to find out what might happened: cell consumed, cell generated, transaction generated, or a chain fork happened, etc.


packages/indexer/README.md
```diff
@@ -171,6 +171,47 @@ for await (const tx of txCollector.collect()) {
...
+    args: "0xa528f2b9a51118b193178db4cf2f3db92e7df323",
+  },
+});
+
+eventEmitter.on("cell_generated",  (outPointBuffer) => {
+    console.log(outPointBuffer);
+    // do something according to the event
+})
+
+eventEmitter.on("chain_forked", (block_number, block_hash) => {
```
xxuejie: I don't think we should provide different events here. We should only provide one event `changed`. It should not come with any data, so as to encourage developers to pull from the data source by themselves. Providing any sort of data might make developers rely on the data here, which contradicts to our design of an unreliable notification queue.

packages/indexer/README.md
```diff
@@ -171,6 +171,47 @@ for await (const tx of txCollector.collect()) {
...
+eventEmitter = indexer.subscribeForTransaction({
+  lock: {
+    code_hash:
+      "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
+    hash_type: "type",
+    args: "0xa528f2b9a51118b193178db4cf2f3db92e7df323",
+  },
+});
+
+eventEmitter.on("transaction_generated",  (txHash) => {
```
xxuejie: Same thing here, a transaction might be committed or reverted(in the case of forks), all we need is a `changed` event with no data.

packages/indexer/native/src/lib.rs
```diff
@@ -110,6 +122,92 @@ impl NativeIndexer {
...
     }
+
+    fn publish_new_block_info(&self, block: &CoreBlockView) {
+        let emitters = self.emitters.read().unwrap();
+        let lock_script_emitters = emitters.iter().filter(|x| x.lock.is_some());
+        let type_script_emitters = emitters.iter().filter(|x| x.type_.is_some());
+
+        let transactions = block.transactions();
+        for (_tx_index, tx) in transactions.iter().enumerate() {
+            // TODO: cell_consumed event ?
```
xxuejie: This is production code, unless absolutely necessary, we shouldn't leave any TODOs.

packages/indexer/native/src/lib.rs
```diff
@@ -110,6 +122,92 @@ impl NativeIndexer {
...
+                                self.emit_transaction_generated_event(emitter, tx.hash());
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    fn publish_fork_info(&self, tip_number: u64, tip_hash: Byte32) {
```
xxuejie: A rollback event is just the reverse of a normal block: you revert all generated cells, and commit all consumed cells. We should test against the cells in the block, and generate the same cell notification events.

xxuejie (changes requested):

packages/indexer/README.md
bitrocks: I have updated the event related logic, but there's one thing to make clear: if we only emit `changed` event with no specific data, what's the different between `subscribeForCell` and `subscribeForTransaction`. As in  https://github.com/nervosnetwork/lumos/issues/55, the api suggests the EventEmitter contains topic related data(`Cell|Transaction`).

packages/indexer/native/src/lib.rs
bitrocks: fixed.

packages/indexer/README.md
xxuejie: You are right here, we shouldn't need 2 APIs, a single subscribe should be enough for both subscribing to cells, and subscribing to transactions. We can just only keep one unified `subscribe` here.

packages/indexer/README.md
bitrocks: updated.

packages/indexer/native/src/lib.rs
bitrocks: updated.

.gitignore
```diff
@@ -109,3 +109,6 @@ packages/sql-indexer/knexfile.dev.js
 # compiled js
 packages/helpers/lib
 packages/common-scripts/lib
+
+# knex config
+packages/sql-indexer/knexfile.ts
```
xxuejie: Please handle proper end of line in a file.

packages/indexer/lib/index.js
```diff
@@ -91,6 +101,23 @@ class Indexer {
   collector({ lock = null, type = null, argsLen = -1, data = "0x" } = {}) {
     return new CellCollector(this, { lock, type, argsLen, data });
   }
+
+  subscribe({ lock = null, type = null } = {}) {
```
xxuejie: Is there any blocker from supporting the full [QueryOptions](https://github.com/nervosnetwork/lumos/blob/develop/packages/base/index.d.ts#L85) interface? I do believe we can support full QueryOptions interface if possible.

bitrocks: Yes, it's feasible to support range parameters(`fromBlock|toBlock`) and args prefix parameters(`argsLen|data`), and the `skip` parameter doesn't make any sense here. But I doubt about the actual use cases of those fine-grained features when monitoring a script's update, and try to avoid adding extra complexity if it's unnecessary.

.gitignore
bitrocks: Oops, I'll fix it.

packages/indexer/lib/index.js
xxuejie: We could ignore those who does not make sense, like `skip`, but I do believe `argsLen`, `data`, and arguable `fromBlock` can be quite helpful for building queries. For example, supporting `multisig` script will naturally need `argsLen`, while Nervos DAO monitoring will require querying `data`. So I do think we should use full `QueryOptions` here.

bitrocks: Ok, I'll add support for `argsLen|data|fromBlock` in the current version.

packages/indexer/lib/index.js
```diff
@@ -88,9 +102,32 @@ class Indexer {
     }, this.livenessCheckIntervalSeconds * 1000);
   }
 
-  collector({ lock = null, type = null, argsLen = -1, data = "0x" } = {}) {
+  collector({ lock = null, type = null, argsLen = -1, data = "any" } = {}) {
```
xxuejie, line 105: With this change, the version released here should be 0.7.0, not 0.6.1

packages/indexer/lib/index.js
```diff
@@ -88,9 +102,32 @@ class Indexer {
...
     return new CellCollector(this, { lock, type, argsLen, data });
   }
+
+  subscribe({
+    lock = null,
+    type = null,
+    argsLen = -1,
+    data = "any",
+    fromBlock = null,
+  } = {}) {
```
xxuejie, line 117: I would suggest we do a little more here: here we are accepting all fields included in QueryOptions. When a developer is passing in a value that subscribe won't use, such as `toBlock` or `skip`, we shall generate a log message saying the passed field is ignored.

packages/indexer/lib/index.js
bitrocks: updated.

packages/indexer/lib/index.js
bitrocks: updated.

xxuejie (approved):

PR #74, opened 2020-07-31
Refactor common scripts

classicalliu:
1. Add `CellCollector` to `secp256k1Blake160`, `secp256k1Blake160Multisig`
2. Refactor `setupInputCell` of `secp256k1Blake160`, `secp256k1Blake160Multisig`, `locktimePool`, `common`
3. Refactor `sudt` to use new `CellCollector`
4. Refactor `injectCapacity` and `transfer` in `common`, can accept customized lock script now.

### DIFF
--- a/packages/common-scripts/src/anyone_can_pay.ts
+++ b/packages/common-scripts/src/anyone_can_pay.ts
@@ -74,11 +74,11 @@ export class CellCollector {
   }
 }
 
-export async function setupInputCell(
+export function setupInputCell(
   txSkeleton: TransactionSkeletonType,
   inputIndex: number,
   { config = undefined }: Options = {}
-): Promise<TransactionSkeletonType> {
+): TransactionSkeletonType {
   config = config || getConfig();
 
   if (inputIndex >= txSkeleton.get("inputs").size) {
--- a/packages/common-scripts/src/anyone_can_pay.ts
+++ b/packages/common-scripts/src/anyone_can_pay.ts
@@ -32,7 +32,7 @@ import { Set, List } from "immutable";
 import { FromInfo, parseFromInfo } from "./secp256k1_blake160_multisig";
 const { CKBHasher, ckbHash, readBigUInt128LE } = utils;
 
-export class CellCollector {
+export class CellCollector implements CellCollectorType {
   private cellCollector: CellCollectorType;
   private config: Config;
   public readonly fromScript: Script;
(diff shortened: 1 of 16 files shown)


packages/common-scripts/src/anyone_can_pay.ts
```diff
@@ -74,23 +76,46 @@ export class CellCollector {
   }
 }
 
-export async function setupInputCell(
+export function setupInputCell(
```
xxuejie: Even though we might not be using it, let's keep the API interface of setupInputCell an async function for potential needs.

packages/common-scripts/src/common.ts
```diff
@@ -1,142 +1,338 @@
...
+      txSkeleton: TransactionSkeletonType,
+      inputCell: Cell,
+      fromInfo: FromInfo,
+      options: {
+        config?: Config;
+        defaultWitness?: HexString;
+        since?: PackedSince;
+        needCapacity?: HexString;
+      }
+    ): {
```
xxuejie: Let's return a promise here for potentail usage

packages/common-scripts/src/common.ts
```diff
@@ -1,142 +1,338 @@
...
-  tipHeader: Header | undefined,
+  changeAddress?: Address,
+  tipHeader?: Header,
   {
-    config,
-    requireToAddress,
-    useLocktimeCellsFirst,
+    config = undefined,
+    useLocktimeCellsFirst = true,
+    customLockScriptInfos = [],
```
xxuejie: I was thinking maybe we can keep an array in this module, and provide a `registerCustomLockScriptInfo` method which puts custom lockScriptInfo into this global array. Notice that we are building a framework(we are already injecting current config to a global variable anyway), this way developers do not have to pass-in customLockScriptInfos every time they innvoke `transfer`.

xxuejie: In the above design, we also can remove the distinction between predefined lockScriptInfos and customLockScriptsInfos. Predefined ones are just lockScriptInfos that is automatically inserted into the global array.

packages/common-scripts/src/anyone_can_pay.ts
classicalliu: updated.

packages/common-scripts/src/common.ts
classicalliu: updated.

packages/common-scripts/src/common.ts
classicalliu: updated.

xxuejie (approved):

PR #76, opened 2020-08-03
fix: Fix typo and object outpoints intersection error

### LINKED ISSUE #75: Typo in Indexer lib leads to failure when attempting to collect cells
typeOoutPoints -> typeOutPoints

Changing the name seems to fix the error. However, I am still not seeing any cells returned when getting by type script only, but that could still be an issue on my end.

https://github.com/nervosnetwork/lumos/blob/81de436ab48f895e8a139200d6565c02d1bbfca5/packages/indexer/lib/index.js#L185

image

bitrocks:
Fix #75, the OrderedSet's intersect function can't perform object level comparison, so it will always return a empty result.

### DIFF
--- a/packages/indexer/lib/index.js
+++ b/packages/indexer/lib/index.js
@@ -2,7 +2,6 @@ const { validators, normalizers, Reader, RPC } = require("ckb-js-toolkit");
 const { OrderedSet } = require("immutable");
 const XXHash = require("xxhash");
 const { Indexer: NativeIndexer } = require("../native");
-const lodash = require("lodash");
 function defaultLogger(level, message) {
   console.log(`[${level}] ${message}`);
 }
@@ -161,6 +160,7 @@ class CellCollector {
           )
           .collect(returnRawBuffer)
       );
+      lockOutPoints = this.wrapOutPoints(lockOutPoints);
     }
 
     if (this.type) {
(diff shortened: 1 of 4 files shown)


packages/indexer/lib/index.js
```diff
@@ -2,7 +2,7 @@ const { validators, normalizers, Reader, RPC } = require("ckb-js-toolkit");
 const { OrderedSet } = require("immutable");
 const XXHash = require("xxhash");
 const { Indexer: NativeIndexer } = require("../native");
-
+const lodash = require("lodash");
```
classicalliu: Where is the `lodash` from? I can't find in `packcage.json`.

bitrocks: Oops, it's in the workspace's `yarn.lock` file, I gonna add it in the package.json.

packages/indexer/lib/index.js
```diff
@@ -196,6 +194,19 @@ class CellCollector {
...
+    return new OrderedSet(
+      lodash.intersectionWith(
+        [...lockOutPoints],
+        [...typeOutPoints],
+        this.isOutPointEqual
+      )
+    );
+  }
+  isOutPointEqual(a, b) {
+    return a.tx_hash == b.tx_hash && a.index == b.index;
```
classicalliu:
```suggestion
    return a.tx_hash === b.tx_hash && a.index === b.index;
```

bitrocks: updated.

xxuejie:
If the problem is just OutPoint equality check, existing [OutPointValue](https://github.com/nervosnetwork/lumos/blob/develop/packages/base/lib/values.js#L39) already does the job with [OrderedSet](https://immutable-js.github.io/immutable-js/docs/#/is), I don't see the need for lodash.

xxuejie (approved):

PR #91, opened 2020-08-19
refactor: Rename `usedCapacity` to `availableCapacity`

classicalliu:


xxuejie (approved):
Notice this is a breaking change, we will need to bump the major version.

PR #98, opened 2020-09-08
feat: Indexer support query with order

bitrocks:


packages/indexer/lib/index.js
```diff
@@ -194,6 +195,9 @@ class CellCollector {
     if (type && typeof type === "object") {
       validators.ValidateScript(type);
     }
+    if (order !== "asc" && order != "desc") {
```
classicalliu:
```suggestion
    if (order !== "asc" && order !== "desc") {
```

packages/indexer/lib/index.js
```diff
@@ -329,12 +339,17 @@ class TransactionCollector {
       this.type = type;
     }
 
+    if (order !== "asc" && order != "desc") {
```
classicalliu:
```suggestion
    if (order !== "asc" && order !== "desc") {
```

packages/sql-indexer/lib/index.js
```diff
@@ -580,6 +581,9 @@ class CellCollector {
     if (type && typeof type === "object") {
       validators.ValidateScript(type);
     }
+    if (order !== "asc" && order != "desc") {
```
classicalliu:
```suggestion
    if (order !== "asc" && order !== "desc") {
```

packages/indexer/lib/index.js
```diff
@@ -249,7 +254,11 @@ class CellCollector {
     } else {
       outPoints = typeOutPoints;
     }
-    return outPoints;
+    if (this.order === "desc") {
+      return outPoints.reverse();
```
xxuejie: Are we sure this is enough? Don't we need to do a query from the other direction at rocksdb side? I recall there are a lot of code to support order query in ckb-indexer

bitrocks: I first came out with the same way you suggesting here, which would query the rocksdb with order parameter via rust bindings directly, but it seems quite ok to reserve the default query results, which is already well-ordered by block number  in ascending order.

bitrocks: Yep, it seems `not` ok after further discussion. gonna fix it.

xxuejie (approved):

PR #99, opened 2020-09-15
Implement hd

classicalliu:


xxuejie (approved):
We should add examples on how to import keystore file from neuron and ckb-cli respectively, then we can merge this.

classicalliu:
Since `Neuron` and `ckb-cli` using the same keystore format, and using `Keystore.load()` can resolve all of these keystore format. Will merge this PR directly.

PR #100, opened 2020-09-16
feat: Change fromBlock, toBlock type from number to Hexadecimal

bitrocks:
BREAKING CHANGES: `fromBlock|toBlock`'s type changed from `number` to `hexString`

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ckb-lumos/base",
-  "version": "0.12.0",
+  "version": "0.13.0-rc1",
   "description": "Base data structures and utilities used in lumos",
   "author": "Xuejie Xiao <xxuejie@gmail.com>",
   "homepage": "https://github.com/nervosnetwork/lumos#readme",
--- a/packages/indexer/lib/index.js
+++ b/packages/indexer/lib/index.js
@@ -137,6 +137,11 @@ class Indexer {
   } = {}) {
     let script = null;
     let scriptType = null;
+    if (fromBlock && !fromBlock.startsWith("0x")) {
+      throw new Error(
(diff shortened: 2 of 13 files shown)


packages/indexer/lib/index.js
```diff
@@ -137,6 +137,11 @@ class Indexer {
   } = {}) {
     let script = null;
     let scriptType = null;
+    if (fromBlock && !fromBlock.startsWith("0x")) {
```
xxuejie: 1. It's better if we test fromBlock is a string first, while TS handles the case when you are writing in TS, there is also the case where people use lumos from plain JavaScript
2. It might be better to split [this part](https://github.com/nervosnetwork/lumos/pull/99/files#diff-36c4b31e9f7cbb1fe191bfb2855bf31cR85) into a separate PR so the code can be shared.

packages/base/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ckb-lumos/base",
-  "version": "0.12.0",
+  "version": "0.12.1",
```
xxuejie: For a breaking change, we need to bump the major version.

packages/indexer/lib/index.js
bitrocks: 1. Add  `assertHexadecimal` util to check the value here;
2. Also add `assertHexString` util in this repo, and follow pr just need to rebase once this one is merged into the develop branch.

packages/base/package.json
bitrocks: I think it would be ok to bump the major version before next release.

bitrocks: Bumped it to v0.13.0-rc1  :)

xxuejie (approved):

PR #106, opened 2020-09-23
feat: Indexer support any argsLen query for transaction collector

bitrocks:


packages/indexer/README.md
```diff
@@ -149,7 +149,7 @@ for await (const cell of cellCollector.collect()) {
 }
 ```
 
-You can also set it as `any` when the argsLen of the `args` might have multiple possibilities. However, there's some performance lost when use `any` rather than explicit specify length due to the low-level implementation.
+ You can also set it as any when the argsLen of the field args might have multiple possibilities, for example, lock script's args could be 20 in normal scenario and 28 in multisig scenario,  or any other length in customized scenarios. However, there's some performance lost when use `any` rather than explicit specify length due to the low-level implementation.
```
classicalliu:
```suggestion
You can also set it as any when the argsLen of the field args might have multiple possibilities, for example, lock script's args could be 20 in normal scenario and 28 in multisig scenario,  or any other length in customized scenarios. However, there's some performance lost when use `any` rather than explicit specify length due to the low-level implementation.
```

packages/indexer/README.md
```diff
@@ -211,32 +240,27 @@ for await (const tx of txCollector.collect()) {
 
 It will fetch transactions between `[fromBlock, toBlock]`, which means both `fromBlock` and `toBlock` are included in query range.
 
-Fine grained query for transactions by scripts with `ioType` is supported: 
+Page jump when queryring transactions is supported:
```
classicalliu:
```suggestion
Page jump when querying transactions is supported:
```

packages/indexer/lib/index.js
```diff
@@ -197,11 +199,28 @@ class CellCollector {
...
+    } else if (lock && lock.script) {
+      validators.ValidateScript(lock.script);
+      this.lock = lock;
+      // check argsLen
+      if (!lock.argsLen) {
+        this.lock.argsLen = argsLen;
+      }
     }
-    if (type && typeof type === "object") {
+    if (type && !type.script) {
```
classicalliu: How about if type equals `empty` as a string?

packages/indexer/lib/index.js
```diff
@@ -197,11 +199,28 @@ class CellCollector {
     if (!lock && typeof type !== "object") {
       throw new Error("Either lock or type script must be provided!");
     }
-    if (lock) {
+    // Wrap the plain `Script` to `ScriptWrapper`.
+    if (lock && !lock.script) {
       validators.ValidateScript(lock);
+      this.lock = { script: lock, argsLen: argsLen };
```
classicalliu, line 205: Missing `ioType` or it has default value ?

packages/indexer/lib/index.js
bitrocks: fixed, add more checks on type option.

packages/indexer/lib/index.js
bitrocks: A code reuse issue, `ioType` in  [line 216](https://github.com/nervosnetwork/lumos/pull/106/files/f714e671fd4ffb34362cf4bff27e95c2fa9388dd#diff-bf9cee66d42f8a2eb8721be2b5937cbeR216) is supposed to be ignored. Fixed.

packages/sql-indexer/lib/index.js
```diff
@@ -577,14 +577,33 @@ class CellCollector {
...
+    } else if (lock && lock.script) {
+      validators.ValidateScript(lock.script);
+      this.lock = lock;
+      // check argsLen
+      if (!lock.argsLen) {
+        this.lock.argsLen = argsLen;
+      }
     }
-    if (type && typeof type === "object") {
+    if (type && type === "empty") {
```
classicalliu:
```suggestion
    if (type === "empty") {
```

classicalliu (approved):

PR #117, opened 2020-10-16
fix: Locktime pool CellCollector no dao cells & add tests

classicalliu:


packages/common-scripts/tests/locktime_pool.test.ts
```diff
@@ -435,3 +449,402 @@ test("Don't update capacity directly when deduct", async (t) => {
...
+          block_hash: bobSecpDaoDepositInput.block_hash,
+        },
+      };
+    }
+
+    throw new Error("Error transaction hash!");
+  }
+}
+
+function cloneObject<T>(obj: T): T {
```
bitrocks, line 532: It's better to use [lodash's cloneDeep](https://lodash.com/docs/4.17.15#cloneDeep) method than the `JSON.parse/stringify` workaround for efficiency and safety reasons, eg. `JSON.parse/stringify` would cause data lost for certain properties  like function, Dates,  `undefined`(exists in the test data), etc.

classicalliu: Replaced with `Immutable.fromJS(obj).toJS()` .

classicalliu: Reverted to using `JSON.parse/stringify`, it's faster than `Immutable.fromJS(obj).toJS()`.
The main purpose for deep clone is cloning nested object, will not include functions or Dates, and key of undefined value lost is unimportant.

bitrocks (approved):

PR #119, opened 2020-10-28
feat: add sql-indexer test code

bitrocks:


packages/sql-indexer/lib/index.js
```diff
@@ -536,6 +540,11 @@ class Indexer {
     toBlock = null,
     skip = null,
   } = {}) {
+    if (lock && type) {
```
classicalliu, line 543: How about `lock` is a script and `type="empty"` ?

bitrocks: The current implementation of notification mechanism doesn't support such subscription yet.

classicalliu (approved):

packages/sql-indexer/lib/index.js
classicalliu: OK, I see.

PR #131, opened 2020-12-17
Median time emitter

classicalliu:


packages/indexer/native/src/lib.rs
```diff
@@ -586,6 +608,19 @@ declare_types! {
...
+
+        method getBlockEmitter(mut cx) {
+            let mut this = cx.this();
+            let block_emitter = JsBlockEmitter::new::<_, JsBlockEmitter, _>(&mut cx, vec![])?;
+            {
+                let guard = cx.lock();
+                let block_emitter = block_emitter.borrow(&guard);
+                let native_indexer = this.borrow_mut(&guard);
+                let mut block_emitters = native_indexer.block_emitters.write().unwrap();
+                block_emitters.push(block_emitter.clone());
```
xxuejie, line 620: I think we only need one block emitter, each median time emitter can subscribe from this single block emitter.

packages/indexer/lib/index.js
```diff
@@ -1,17 +1,133 @@
...
 
+class EventMap {
+  constructor() {
+    this.map = Map();
+  }
+
+  add(key, value) {
+    const v = this.map.get(key);
+    if (v === undefined) {
+      this.map = this.map.set(key, {
```
xxuejie: All you need here is a plain Map, you can do:

```
const map = Map();

map[v] = map[v] || 0;
map[v] = map[v] + 1;
```

xxuejie (approved):

PR #148, opened 2021-01-15
Add rpc module

classicalliu:
* Add a new `rpc` module, provides type definitions for rpc requests and response.
* Optional `waitForSync` after every rpc call in new `rpc` module.
* Replace `common-scripts` & `hd-cacahe`'s `rpc` module with new `rpc` module.


packages/rpc/src/index.ts
```diff
@@ -0,0 +1,358 @@
...
+  }
+}
+
+export class RPC {
+  private rpcProxy: any;
+
+  /**
+   *
+   * @param uri
+   * @param indexer Will `waitForSync` after rpc call if provided.
```
xxuejie, line 107: Is there a way to opt out of the waitForSync behavior?

classicalliu: How about add a new method?

```typescript
resetIndexer(indexer?: Indexer) {
  this.indexer = indexer
}
```

Or you mean exit from a running `waitForSync` ?

xxuejie: I just mean that there should a way to disable the `waitForSync` behavior. For the average crowd we want to keep them on the safe side, but for the dangerous tweakers they should be able to kill this feature for more performance.

classicalliu: A `resetIndexer` method added.

https://github.com/nervosnetwork/lumos/pull/148/commits/1b345dea9045555f69503b7e2b185f0f290722bf

xxuejie (approved):

PR #151, opened 2021-02-02
feat: add rpcOptions to Sql Indexer constructor

bitrocks:


packages/rpc/src/index.ts
```diff
@@ -59,12 +59,14 @@ class RpcProxy {
     {
       waitForSyncCheckIntervalSeconds = 1,
       blockDifference = 3,
+      rpcOptions = {},
     }: {
       waitForSyncCheckIntervalSeconds?: number;
       blockDifference?: number;
+      rpcOptions?: object;
```
classicalliu, line 66: Also to `RPC` class, it's the exported interface.

classicalliu (approved):

PR #158, opened 2021-07-27
(DONT MERGE)feat: update getBlockEmitter impl

bitrocks:


xxuejie:
Closing this, the native indexer in lumos will be deprecated. When mercury v0.2.0 launches, lumos will leverage mercury as the indexing solution instead.

PR #174, opened 2021-10-22
feat: add deprecated annotation to indexer and sql-indexer

sighWang:


sighWang:

PR #175, opened 2021-10-22
feat: add ckb-indexer and ckb-collector

sighWang:
will add test next pr


homura:
is this duplicated with #178? if so maybe we should close this PR

sighWang:

PR #176, opened 2021-10-25
feat(ckb-indexer):  can use cell collector whit skip option

sighWang:


sighWang:

PR #179, opened 2021-10-28
feat: supported new full address format

homura:
This PR is adapted to CKB 2021 hardfork, it works for

- [supported `extra_hash`](https://github.com/nervosnetwork/ckb/pull/2776)
- [supported VM version selection](https://github.com/nervosnetwork/rfcs/blob/c0aacf775a347ec44b3b0c02ba6dd67a5a47128c/rfcs/0032-ckb-vm-version-selection/0032-ckb-vm-version-selection.md)
- [new full address format](https://github.com/nervosnetwork/rfcs/pull/239)

This PR is a supplement to #172

### DIFF
--- a/packages/config-manager/lib/predefined.js
+++ b/packages/config-manager/lib/predefined.js
@@ -108,23 +108,7 @@ const AGGRON4 = deepFreeze({
   },
 });
 
-function CKB2019(config) {
-  return {
-    ...config,
-    CKB2021: false,
-  };
-}
-
-function CKB2021(config) {
-  return {
-    ...config,
-    CKB2021: true,
-  };
-}
-
 module.exports = {
   LINA,
   AGGRON4,
-  CKB2019,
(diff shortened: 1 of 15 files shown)


packages/config-manager/lib/predefined.js
```diff
@@ -1,6 +1,7 @@
 const deepFreeze = require("deep-freeze-strict");
 
 const LINA = deepFreeze({
+  CKB2021: true,
```
cryptowen, line 4: Should this be `false`?

homura: The latest version [0.18.x](https://github.com/nervosnetwork/lumos/tree/ckb2021) would only support CKB2021, the default config should be 2021 here. The CKB2021 config leaves here only affect address generation. We're not sure if all the tools in the community are using the new address format, but we've left open the possibility of using legacy addresses

PR #181, opened 2021-11-04
feat: make config-manager works with browser

homura:
- rewrite in TypeScript lang
- make it autocomplete friendly
- works with browser

image


homura:
Usage can be found in the [example](https://github.com/nervosnetwork/lumos/blob/develop/examples/config-manager.ts)

PR #185, opened 2021-11-05
feat(ckb-indexer): can get cell with block hash

sighWang:
* copy plimit and queue file cause es-module import error
* ignore and add todo to some ts type warning


packages/ckb-indexer/src/pLimit.ts
```diff
@@ -0,0 +1,86 @@
+import Queue from "./queue";
```
homura: to make the source trackable, maybe we should move the file to `ckb-index/src/third-party/*`, and add a comment about the source's GitHub URL

homura:
we should add comments to `CKBIndexer` and `CKBCellCollector` to describe how `withBlockHash` works, and we should also add a migration doc to describe how to migrate from `NativeIndexer` to `CKBIndexer`

packages/ckb-indexer/src/collector.ts
```diff
@@ -234,10 +241,33 @@ export class IndexerCollector implements BaseCellCollector {
     return counter;
   }
 
+  private async getLiveCellWithBlockHash(lastCursor?: string) {
+    let result: GetCellsResults = await this.getLiveCell(lastCursor);
+    const limit = pLimit(10);
+    //TODO refine this functions type
+    const inputs = result.objects.map((cell) => {
```
homura: found the JSON-RPC is [batchable](https://www.jsonrpc.org/specification#batch)，maybe we can implement via the batch spec.

sighWang: use json-rpc batch and test success both local and testnet rpc server

packages/ckb-indexer/src/collector.ts
```diff
@@ -12,13 +12,26 @@ import {
   GetCellsResults,
   HexadecimalRange,
   Order,
+  rpcResponse,
   SearchFilter,
 } from "./indexer";
 
 import { CkbIndexer, ScriptType, SearchKey } from "./indexer";
+import axios from "axios";
```
homura: axios is too heavy, we can switch to cross-fetch

packages/ckb-indexer/src/collector.ts
```diff
@@ -234,10 +245,64 @@ export class IndexerCollector implements BaseCellCollector {
...
+    }
+    const requestData = result.objects.map((cell, index) => {
+      return {
+        id: index,
+        jsonrpc: "2.0",
+        method: "get_block_hash",
+        params: [cell.block_number],
+      };
+    });
+    const blockHashList: any[] = await this.request(
```
```suggestion
    // TODO define type here
    const blockHashList: any[] = await this.request(
```

packages/ckb-indexer/src/collector.ts
sighWang: refactor(ckb-indexer): use cross-fetch instand of axios

packages/ckb-indexer/src/collector.ts
sighWang: feat(ckb-indexer): add type to blockHash list

packages/ckb-indexer/src/collector.ts
```diff
@@ -24,6 +24,12 @@ export interface OtherQueryOptions {
   ckbRpcUrl: string;
 }
 
+export interface RPCResult {
```
```suggestion
interface GetBlockHashRPCResult {
```

sighWang: refactor(ckb-indexer): rename

PR #191, opened 2021-11-09
deploy generator

liubq7:


packages/base/lib/utils.js
```diff
@@ -96,6 +96,28 @@ function assertHexadecimal(debugPath, str) {
...
+function generateTypeIdArgs(input, outputIndex) {
+  const outPointBuf = SerializeCellInput(normalizers.NormalizeCellInput(input));
+  const outputIndexBuf = toBigUInt64LE(outputIndex);
+  const ckbHasher = new CKBHasher();
+  ckbHasher.update(outPointBuf);
+  ckbHasher.update(outputIndexBuf);
+  return ckbHasher.digestHex();
+}
+
+function generateTypeIdScript(input, outputIndex) {
```
homura: we should check if input params are valid

```ts
assertsValidScript(input);
assertsValidHexNumber(outputIndex);
```

packages/common-scripts/src/deploy.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  txSkeleton = updateOutputs(txSkeleton, output);
+  txSkeleton = await completeTx(
+    txSkeleton,
+    fromAddress,
+    options.feeRate,
+    options.config
+  );
+
+  return updateCellDeps(txSkeleton, options.config);
```
homura: `updateCellDeps` after `payFee` may cause wrong fee calculation since the tx size has changed.  the deploy generator implemented here can refer to the implementation of [commons.transfer](https://github.com/nervosnetwork/lumos/blob/519ddf213b6164442e0e1fcdcbbb7d4b96456bed/packages/common-scripts/README.md), it return a `txSkeleton` that unpaid for fee

packages/common-scripts/src/deploy.ts
```diff
@@ -0,0 +1,275 @@
...
+/**
+ * Generate txSkeleton for writing binary data to CKB via Type ID, usually for deploying contracts.
+ * Deploying via Type ID makes it possible to upgrade contract, for more information about Type ID, please check: https://xuejie.space/2020_02_03_introduction_to_ckb_script_programming_type_id/
+ * This generator only supports `SECP256K1_BLAKE160` and `SECP256K1_BLAKE160_MULTISIG` currently.
+ *
+ * @param options
+ */
+export async function generateDeployWithTypeIdTx(
+  options: DeployOptions
+): Promise<TransactionSkeletonType> {
```
homura: maybe we also need to return the type_id script from this generator, 

```ts
export async function generateDeployWithTypeIdTx(
  options: DeployOptions
): Promise<[Script /* type_id script */, TransactionSkeletonType]>
```

packages/base/lib/utils.js
homura:
```suggestion
function generateTypeIdScript(input, outputIndex = '0x0') {
```

packages/base/lib/utils.js
```diff
@@ -96,6 +96,33 @@ function assertHexadecimal(debugPath, str) {
...
+  validators.ValidateCellInput(input);
+  outputIndex = outputIndex === null ? "0x0" : outputIndex;
```
homura: the `outputIndex === null` only matched when `outputIndex` IS actually `null`, maybe the `outputIndex == null` is better or we can just use `0x0` as  default parameter for `outputIndex`

packages/base/lib/utils.js
```diff
@@ -96,6 +96,33 @@ function assertHexadecimal(debugPath, str) {
...
+  const re = /^0x[0-9a-fA-F]/;
+  if (!re.test(outputIndex)) throw new Error("invalid outputIndex");
```
homura: the `assertHexadecimal` is just above

packages/common-scripts/README.md
```diff
@@ -155,3 +157,54 @@ txSkeleton = await sudt.transfer(
...
+```javascript
+const { generateDeployWithDataTx, generateDeployWithTypeIdTx, generateUpgradeTypeIdDataTx, payFee } = require("@ckb-lumos/common-scripts");
+const { Indexer } = require("@ckb-lumos/ckb-indexer");
+const { initializeConfig, predefined } = require("@ckb-lumos/config-manager");
+const { parseAddress } = require("@ckb-lumos/helpers");
+
+initializeConfig(predefined.AGGRON4);
+
+const CKB_RPC_URL = "https://testnet.ckb.dev/rpc";
+const CKB_INDEXER_URL = "https://testnet.ckb.dev/indexer";
```
homura:
```suggestion
const CKB_RPC_URL = "http://localhost:8114";
const CKB_INDEXER_URL = "http://localhost:8116";
```

it is safer to maintain our own nodes than to rely directly on public nodes

packages/common-scripts/src/deploy.ts
```diff
@@ -0,0 +1,415 @@
...
+  });
+  return txSkeleton;
+}
+
+async function injectCapacity(
+  txSkeleton: TransactionSkeletonType,
+  fromAddress: string,
+  amount: bigint,
+  { config = undefined }: Options = {}
+): Promise<TransactionSkeletonType> {
```
homura: since we should support the SECP256K1 and SECP256K1_MULTISIG, i found that the implementation of `injectCapacity` here seems very similar to secp256k1's `transfer`, which means that the `injectCapacity` only support `SECP256K1_BLAKE160` but not `MULTI_SIG`, can we call the functions provided by secp256k1/multi_sig directly to implement the `injectCapacity` here?

packages/common-scripts/src/deploy.ts
```diff
@@ -0,0 +1,558 @@
...
+  const output: Cell = {
+    cell_output: {
+      capacity: "0x0",
+      lock: options.outputScriptLock,
+      type: options.typeId,
+    },
+    data: bytesToHex(options.scriptBinary),
+  };
+
+  txSkeleton = updateOutputs(txSkeleton, output);
```
homura, line 597: if the contract is reduced in size when updating, will the capacity be consumed by using `updateOutput` directly?

homura:
maybe we should add some unit tests

1. when `collectedCapacity >= output + change cell `
2. when `collectedCapacity >= output` but `collectedCapacity < output + change cell `
3. when `collectedCapacity < output`

we should make sure the tx fee is calculated correctly


```ts
// successful case
test("collected capacity is enough for change cell and deploy cell", () => {
  const dummyProvider = {collector() {collect() { return [...] }}}

  const {tx} = await generateDeployTxWithData(...)

  t.is(tx.get('outputs')[changeCellIndex].get('capacity'), ...)
})

// failed case
test("collected capacity is NOT enough for change cell and deploy cell", () => {
  const dummyProvider = {collector() {collect() { return [...] }}}

  const error = t.throwAsync(() => generateDeployTxWithData(...))

  t.is(error.message, ...)
})

```

packages/common-scripts/src/deploy.ts
```diff
@@ -0,0 +1,572 @@
...
+
+/**
+ * Generate txSkeleton for writing binary data to CKB, usually for deploying contracts.
+ * This generator only supports `SECP256K1_BLAKE160` and `SECP256K1_BLAKE160_MULTISIG` currently.
+ *
+ * @param options
+ */
+export async function generateDeployWithDataTx(
+  options: DeployOptions
+): Promise<DeployResult> {
```
homura, line 482: We should first verify that `FromInfo` is secp256k1 or multisig, and throw an error if it is not

```ts
function isMultisigFromInfo(fromInfo: FromInfo): fromInfo is MultisigScript {
  if(typeof fromInfo !== 'object') return false;
  return "M" in fromInfo && "R" in fromInfo && Array.isArray(fromInfo.publicKeyHashes);
}

if (typeof fromInfo === 'string') {
  if(nameOfScript(parseAddress(fromInfo), config.SCRIPTS) !== 'SECP256K1_BLAKE160') throw new Error('...')
} else if (!isMultisigFromInfo(fromInfo)) {
  throw new Error('only SECP256K1_BLAKE160 or SECP256K1_MULTISIG is supported ')
} 

 
```

packages/common-scripts/src/deploy.ts
```diff
@@ -17,11 +17,13 @@ import {
   minimalCellCapacity,
   Options,
   createTransactionFromSkeleton,
+  parseAddress,
 } from "@ckb-lumos/helpers";
 import { Reader, normalizers } from "ckb-js-toolkit";
 import { RPC } from "@ckb-lumos/rpc";
 import { Set } from "immutable";
 import { FromInfo, parseFromInfo, MultisigScript } from "./from_info";
+import { nameOfScript } from "../../config-manager/src/helpers";
```
homura: Maybe this is a wrong import path

PR #201, opened 2021-11-29
feat(helpers)!: deprecated short fmt address

homura:
### `generateAddress`

```ts
// before
generateAddress(script) 

// after
generateAddress(script, { __generateShortAddressWhenShortIDInConfig: true } )
```

note: `parseAddress` We will be compatible with the current all format

## Background

Due to the upgrade of CKB2021, address format will also be upgrade, but this will cause the address to be difficult to understand,  after CKB2021 hard-fork, it will become `CKB2021 full address` /  `CKB2021 short address` / `CKB2019 full address` / `CKB2019 short address`. In order to avoid the complexity of address format, the CKB community will migrate to only support full format address.

### DIFF
--- a/packages/config-manager/src/types.ts
+++ b/packages/config-manager/src/types.ts
@@ -5,7 +5,11 @@ export interface ScriptConfig {
   TX_HASH: string;
   INDEX: string;
   DEP_TYPE: "dep_group" | "code";
-  /** Short ID for creating CKB address, not all scripts have short IDs. */
+  /**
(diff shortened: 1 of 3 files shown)


homura:
This breaking change may cause the Lumos upgrade still compiles fine, but the address returned is not the same as the old version, so we decided to use a new API instead of `generateAddress` and mark `generateAddress` as deprecated

homura:
We can find the the new API in #205

PR #203, opened 2021-11-30
feat(ckb-indexer): add ckb-indexer rpc module

sighWang:


packages/ckb-indexer/src/rpc.ts
```diff
@@ -0,0 +1,70 @@
...
+    after_cursor?: string
+  ): Promise<GetTransactionsResults> {
+    return this.rpcProxy.get_transactions(
+      searchKey,
+      order,
+      limit,
+      after_cursor
+    );
+  }
+  async get_cells_capacity(searchKey: SearchKey) {
```
homura: missing return type

PR #208, opened 2021-12-05
feat(ckb-indexer): get transaction batch while transaction io type is input

sighWang:


packages/ckb-indexer/tests/transaction_collector-special.test.ts
```diff
@@ -0,0 +1,20 @@
+import { HashType } from "@ckb-lumos/base";
+import test from "ava";
+import { Indexer, TransactionCollector } from "../src";
+
+const nodeUri = "https://testnet.ckb.dev/rpc";
```
homura: we shouldn't test dependent on testnet data

PR #209, opened 2021-12-07
chore: bump version

homura:


homura:
close this since we added lerna as a version manager https://github.com/nervosnetwork/lumos/pull/210

PR #212, opened 2021-12-08
feat(ckb-indexer): migrate bigint to JSBI

sighWang:


packages/ckb-indexer/src/type.ts
```diff
@@ -113,7 +114,7 @@ export class IndexerEmitter extends EventEmitter {
   type?: Script;
   outputData?: HexString | "any";
   argsLen?: number | "any";
-  fromBlock?: bigint;
```
sighWang, line 116: fromBlock return in indexer.subscribe as EventEmitter.fromBlock.
cause eventEmitter only used as eventEmitter.on('eventName', handle)

so I think EventEmitter.fromBlock can be private filed. So we can refactor to  ` fromBlock?: JSBI;`.
otherwise we need a  method named `subscibeCompatible`

@homura

homura:
maybe we should add some migration suggestion in the [doc](https://github.com/sighWang/lumos/blob/ckb2019-jsbi/documentations/migration-from-bigint-to-jsbi.md#related-api)

packages/common-scripts/src/helper.ts
```diff
@@ -143,6 +144,18 @@ export function isAcpAddress(address: Address, config: Config): boolean {
...
+  if (witnessHexString.length <= 8) {
+    view.setUint32(0, Number("0x" + witnessHexString), true);
+    view.setUint32(4, Number("0x" + "00000000"), true);
+  }
+
+  if (witnessHexString.length > 8 && witnessHexString.length <= 16) {
+    view.setUint32(0, Number("0x" + witnessHexString.slice(-8)), true);
+    view.setUint32(4, Number("0x" + witnessHexString.slice(0, -8)), true);
+  }
   view.setBigUint64(0, BigInt(new Reader(witness).length()), true);
```
homura: the `view.setBigUint64`  is still using the BigInt feature here, we should replace the API with `view.setUint32`

sighWang: deleted

sighWang: deleted

PR #225, opened 2021-12-30
feat: sync develop into ckb2019-jsbi

sighWang:


packages/common-scripts/tests/common-bigint.tes.ts
```diff
@@ -0,0 +1,112 @@
+import test from "ava";
```
homura: missed filename `test`

packages/ckb-indexer/src/transaction_collector.ts
```diff
@@ -118,22 +118,24 @@ export class CKBIndexerTransactionCollector extends BaseIndexerModule.Transactio
...
+            const targetTx = transactionList.find(
+              (tx) => tx.transaction.hash === transactionHash
+            );
+            if (targetTx) {
+              targetTx.inputCell = output;
+            }
+          });
         });
-      });
+    }
```
homura, line 138: Is this fix in the develop branch, if not, this fix should be in a separate PR to ensure that all branches have it

yarn.lock
```diff
@@ -3091,6 +3091,13 @@ babel-plugin-polyfill-regenerator@^0.3.0:
   dependencies:
     "@babel/helper-define-polyfill-provider" "^0.3.0"
 
+babel-plugin-transform-bigint@^1.0.12:
+  version "1.0.12"
+  resolved "https://registry.yarnpkg.com/babel-plugin-transform-bigint/-/babel-plugin-transform-bigint-1.0.12.tgz#d00aff7f9b210faf1853283b798a086956a6d01c"
+  integrity sha512-WJ6BM9ERc8RGv/Oj/UYSs3BX6OCsE4Y70qOh+8c8l4SWc+2js3DmHRTGGbdJXMOtj9UnEKgnr1bfSOvw3GcWuQ==
+  dependencies:
+    "@babel/plugin-syntax-bigint" "^7.8.3"
+
```
homura: I think we don't need the dep here

packages/ckb-indexer/src/transaction_collector.ts
sighWang: https://github.com/nervosnetwork/lumos/pull/235

yarn.lock
sighWang: removed in next commit

PR #226, opened 2022-01-06
big number lib for lumos

liubq7:


Makefile
```diff
@@ -21,6 +22,7 @@ test:
 	yarn workspace @ckb-lumos/indexer test
 	yarn workspace @ckb-lumos/transaction-manager test
 	yarn workspace @ckb-lumos/rpc test
+	yarn workspace @ckb-lumos/bi test
```
homura: we should build bi at first since it is a low level lib

packages/bi/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+      return value;
+    } else if (isBIish(value)) {
+      return toBI(toJSBI(value));
+    } else {
+      throw new Error(`invalid type: ${value} can't be converted into BI`);
+    }
+  }
+
+  static isBI(value: any): value is BI {
+    return value instanceof BI;
```
homura: it would be broken when a user uses `bi@0.0.1` and `bi@0.0.2`togther.

```ts
import { f } from 'other-lib'
import {BI} from 'bi' // @0.0.2

const num = f(); /* return a BI@0.0.1 */
BI.isBI(num) // false
```
maybe we should [impl](https://github.com/ethers-io/ethers.js/blob/master/packages/bignumber/src.ts/bignumber.ts#L294) `isBI` in this way

packages/bi/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+
+  toString(radix?: number): string {
+    radix = radix || 10;
+    return this.jsbi.toString(radix);
+  }
+
+  toHexString(): string {
+    return this.jsbi.toString(16);
```
homura:
```suggestion
    return '0x' + this.jsbi.toString(16);
```

homura (approved):

PR #237, opened 2022-01-17
feat(website): add tools page to website 

sighWang:


homura:
It does not work after building the website

```sh
cd docusaurus/website
yarn install
yarn run build
yarn run serve
```

sighWang:
tools use API like helper.encodeToAddress only implement on branch ckb2021,So I create a new PR https://github.com/nervosnetwork/lumos/pull/241  and change merge target to ckb2021. this pr will be close

PR #239, opened 2022-01-18
feat: migrate exposed APIs from JSBI to BI 

sighWang:
We have previously released a library [@ckb-lumos/bi](https://github.com/nervosnetwork/lumos/pull/226) for handling big numbers by wrapping JSBI to avoid the compatibility issues caused by `bigint`. 

This PR uses `@ckb-lumos/bi` to upgrade the lumos API that was previously exposed to the public using `bigint` in a compatible way, see the [migration suggestion](https://github.com/nervosnetwork/lumos/blob/285539de161f4873f424943c7d33fb1bb15310f8/documentations/migration-from-bigint-to-bi.md) for details of how to do this and how to migrate.


homura:
would you please update the [migration guide](https://github.com/nervosnetwork/lumos/blob/ckb2019-jsbi/documentations/migration-from-bigint-to-jsbi.md) in this PR?

sighWang:
> would you please update the [migration guide](https://github.com/nervosnetwork/lumos/blob/ckb2019-jsbi/documentations/migration-from-bigint-to-jsbi.md) in this PR?

updated

packages/ckb-indexer/package.json
```diff
@@ -25,7 +25,10 @@
   ],
   "dependencies": {
     "@ckb-lumos/base": "^0.17.0-rc7",
+    "@ckb-lumos/bi": "^0.17.0-rc7",
     "@ckb-lumos/rpc": "^0.17.0-rc7",
+    "@types/sinon": "^10.0.6",
+    "ava": "^3.15.0",
```
homura:
```suggestion

```

sighWang: removed

PR #245, opened 2022-01-22
chore: add commitlint for commit message

sighWang:
add rules to commit message. check  https://sparkbox.com/foundry/semantic_commit_messages for more detaile.

```
chore: add Oyster build script
docs: explain hat wobble
feat: add beta sequence
fix: remove broken confirmation message
refactor: share logic between 4d3d3d3 and flarhgunnstow
style: convert tabs to spaces
test: ensure Tayne retains clothing
```


commitlint.config.ts
```diff
@@ -0,0 +1,30 @@
...
+  "toolkit",
+  "transaction-manager",
+  "examples",
+  "website",
+];
+const Configuration: UserConfig = {
+  extends: ["@commitlint/config-conventional"],
+  rules: {
+    "scope-enum": [2, "always", scopeEnumValues],
+    "scope-empty": [2, "never"],
```
homura: maybe we should NOT ban an empty scope, just like this commit, `chore: lint for commit message` is just fine

sighWang: In order to prevent some one forget to add scope which should be added, I set scope-empty not allowed.
and for global commit message we can add `lumos` instand.

sighWang: removed scope-empty rule

PR #246, opened 2022-01-23
feat: upgrade exposed API from BigInt to BI

homura:
This PR is a merge PR, for more information, we can check https://github.com/nervosnetwork/lumos/pull/239


homura:
We have migrated to another JSBI wrapped implementation `@ckb-lumos/bi` https://github.com/nervosnetwork/lumos/pull/248 so close this one

PR #247, opened 2022-01-24
feat(website):add pw lock ,cheque and omni lock named lock

sighWang:


docusaurus/website/src/components/AddressToScript.tsx
```diff
@@ -30,6 +30,79 @@ interface ModalFormValues {
   address: string;
 }
 
+const LINA_SCRIPTS = {
```
homura:
```suggestion
const LINA_SCRIPTS: ScriptRecord = {
```

we can explicitly tell the TS compiler the type of an object to avoid writing `'type' as 'data' | 'type'`

sighWang: use `config.ScriptConfigs` instand of `type' as 'data' | 'type'`

PR #248, opened 2022-01-25
feat: upgrade exposed API from BigInt to BI

sighWang:
We have previously released a library [@ckb-lumos/bi](https://github.com/nervosnetwork/lumos/pull/226) for handling big numbers by wrapping JSBI to avoid the compatibility issues caused by `bigint`. 

This PR uses `@ckb-lumos/bi` to upgrade the lumos API that was previously exposed to the public using `bigint` in a compatible way, see the [migration suggestion](https://github.com/nervosnetwork/lumos/blob/285539de161f4873f424943c7d33fb1bb15310f8/documentations/migration-from-bigint-to-bi.md) for details of how to do this and how to migrate.


packages/base/lib/index.js
```diff
@@ -7,4 +9,6 @@ module.exports = {
   helpers: require("./helpers"),
   indexer: require("./indexer"),
   logger: require("./logger"),
+  primitive,
+  ...primitive,
```
homura: we should not expose primitive since JSBI is NOT a primitive anymore

packages/base/lib/primitive.d.ts
```diff
@@ -1,17 +1,23 @@
+import primitive from "jsbi";
```
homura: can we remove `jsbi` in primitive?

packages/base/lib/primitive.js
```diff
@@ -1 +1,84 @@
 // for typescript
+const JSBI = require("jsbi");
```
homura: can we remove JSBI ?

packages/base/lib/since.js
```diff
@@ -89,98 +99,110 @@ function generateAbsoluteEpochSince({ length, index, number }) {
...
+      if (
+        !BI.from(added.length).eq(0) &&
+        BI.from(added.index).gte(added.length)
+      ) {
+        let _x, _y;
+
+        (_x = added),
+          (_y = "number"),
+          (_x[_y] = BI.from(added.index).div(added.length).add(_x[_y]));
+        added.index = BI.from(added.index).mod(added.length);
```
homura: would you please refactor the `_x` `_y` for more readable

homura:
would you please make a PR to the `ckb2021` branch?

packages/base/lib/primitive.js
sighWang: can we remove primitve.js?
cause there is no code use primitive.js and BI package can do everything primitive.js can.

packages/base/lib/index.js
sighWang: deleted

packages/base/lib/primitive.d.ts
sighWang: deleted

packages/base/lib/primitive.js
sighWang: remove primitive.js code

packages/base/lib/since.js
sighWang: https://github.com/nervosnetwork/lumos/pull/248/commits/9f1b8fdbe6fcc08eef2873a1f1386940736cbeba

PR #254, opened 2022-02-10
Add P2PKH general function

liubq7:
Many witness signing processes are similar to P2PKH signing. Writing a general function to deal with such situations can reduce code repetition.

Check README for more usage details.

## Usage

```ts
import { common } from '@ckb-lumos/lumos';
// import { createP2PKHMessage } from '@ckb-lumos/common-scripts';

const { createP2PKHMessageGroup } = common;

const [{ message }] = createP2PKHMessageGroup(tx, [tx.inputs.get(0)!])
const witness = SerializeWitnessArgs({ lock: new Reader(message) })
...
```


packages/common-scripts/README.md
```diff
@@ -208,3 +210,132 @@ const upgradeOptions = {
...
+      // keccak.update(message);
+    } else {
+      keccak.update(Buffer.from(new Uint8Array(message as ArrayBuffer | Buffer)));
+    }
+  },
+  digest: () => "0x" + keccak.digest("hex"),
+})
+
+// for each Group, sign the message with the lock's corresponding private key, and update witness[index] with the signature.
+```
```
homura: The code in the README is a bit verbose, we can actually modify the code in the [examples](https://github.com/nervosnetwork/lumos/tree/develop/examples) directly and just link to the file in the README

packages/common-scripts/tests/p2pkh.test.ts
```diff
@@ -0,0 +1,512 @@
...
+      HASH_TYPE: "type",
+      TX_HASH:
+        "0x57a62003daeab9d54aa29b944fc3b451213a5ebdf2e232216a3cfed0dde61b38",
+      INDEX: "0x0",
+      DEP_TYPE: "code",
+    },
+  },
+});
+
+test("omni lock [g1]", (t) => {
```
homura, line 9: the test case code is too verbose, can you split the test case into fixtures(maybe represent as a `JSON`) and test logic?

it also helps make it easier to test when other languages also need to implement signature algorithms

```ts

test('test createP2PKHMessage', (t) => {
  const [omni, pw, secp256k1]  = fixtures
  
  expect(createP2PKHMessage(transactionToSkeleton(omni.tx))).is(omni.message);
  // ...
})

```

```json
[{"name": "seck256k1 lock [g1]", "tx": {...}, "result": "0x..."}]
```

packages/common-scripts/src/p2pkh.ts
```diff
@@ -0,0 +1,136 @@
...
+  return utils.ckbHash(
+    core.SerializeRawTransaction(
+      normalizers.NormalizeRawTransaction(createTransactionFromSkeleton(tx))
+    )
+  );
+}
+
+export interface Hasher {
+  update(message: ArrayBuffer): void;
+  digest(): Hash;
```
homura: `Hash` is actually an alias of a string, and we should avoid errors in Hasher's implementers, most of the hasher in the current js ecosystem use `Uint8Array` as the structure for binary data, so it is better to use a definition like this for Hasher

```ts
interface Hasher {
  update(bytes: Uint8Array): void;
  digest(): Uint8Array;
}
```

Usage:

```ts
// with NodeJS
const customHasher: Hasher = require("crypto").createHash("sha256");
```

packages/common-scripts/tests/p2pkh.test.ts
```diff
@@ -0,0 +1,214 @@
...
+  const tmpWitnessArgs = { lock: SECP_SIGNATURE_PLACEHOLDER };
+  const tmpWitness = new Reader(
+    core.SerializeWitnessArgs(normalizers.NormalizeWitnessArgs(tmpWitnessArgs))
+  ).serializeJson();
+  tx = tx.update("witnesses", (witnesses) =>
+    witnesses.push(tmpWitness, tmpWitness)
+  );
+
+  const hasher = new utils.CKBHasher();
+  const signLock = p2pkhJson["OMNI_LOCK_[G1]"].SIGN_LOCK as Script;
```
homura: maybe we should extract this section as a function, and put the `tmpWitness` into the fixture file

```ts
function txSkeletonFromJson(tx: TransactionSkeletonType): object;
function txSkeletonToJson(object: object): TransactionSkeletonType;
```

```ts
const skeleton = txSkeletonFromJson(p2pkhFixtures['some case'].tx);
const signLock = {...}
const [{message}] = createP2PKHMessageGroup(skeleton, {group: [signLock]})
t.is(message, p2pkhFixtures['some case'].signingMessage)
```

packages/common-scripts/src/p2pkh.ts
```diff
@@ -0,0 +1,136 @@
...
+ * @param tx TxSkeleton with all input cells' witnessArgs.lock filled with 0.
+ * @param locks Locks you want to sign, e.g. you don't need to sign ACP cells.
+ * @param hasher Message hasher, defaults to CKB blake2b hasher. Check
+ * https://github.com/nervosnetwork/ckb-system-scripts/blob/e975e8b7d5231fdb1c537b830dd934b305492417/c/secp256k1_blake160_sighash_all.c#L22-L28 for more.
+ * @returns An array of Group containing: lock of the input cell you need to sign, message for signing, witness index of this message (first index of the input cell with this lock).
+ */
+export function createP2PKHMessageGroup(
+  tx: TransactionSkeletonType,
+  locks: Script[],
+  hasher?: Hasher
```
homura: This function may be extended in the future, and the second parameter should preferably be exposed as an object, otherwise, it will probably look like this when the function needs to be extended

```ts
export function createP2PKHMessageGroup(tx, locks, hasher?, feature1?) {}
export function createP2PKHMessageGroup(tx, locks, hasher?, feature1?, feature2?) {}
export function createP2PKHMessageGroup(tx, locks, hasher?, feature1?, feature2?, ...) {}

```

packages/common-scripts/tests/p2pkh.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  let tx = txSkeletonFromJson(
+    p2pkhJson["OMNI_LOCK_[G1]"].tx as txObject,
+    SIGNATURE_PLACEHOLDER
+  );
+
+  const hasher = new utils.CKBHasher();
+  const signLock = p2pkhJson["OMNI_LOCK_[G1]"].SIGN_LOCK as Script;
+
+  const messageGroup = createP2PKHMessageGroup(tx, [signLock], {
+    hasher: {
```
homura, line 20: why not use default hasher here?

liubq7: No reason. I think we are still allowed to pass in default value and it won't affect performance.

PR #267, opened 2022-02-22
chore(examples): cardano-lock integration example

liubq7:
There will be an error after `yarn start`:
unknown

Refresh again and it will work.


examples/cardano-lock-namiwallet/src/lib.ts
```diff
@@ -0,0 +1,208 @@
...
+}
+
+function hashWitness(hasher: utils.CKBHasher, witness: ArrayBuffer): void {
+  const lengthBuffer = new ArrayBuffer(8);
+  const view = new DataView(lengthBuffer);
+  view.setBigUint64(0, BigInt(new toolkit.Reader(witness).length()), true);
+
+  hasher.update(lengthBuffer);
+  hasher.update(witness);
+}
```
homura:
```suggestion
```

PR #277, opened 2022-03-02
test: fix ckb indexer e2e test ckb2021

sighWang:


packages/testkit/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@ckb-lumos/testkit",
-  "version": "0.18.0-rc5",
-  "private": true,
+  "version": "0.17.0-rc8",
+  "private": false,
```
homura, line 4: we should not modify the version and private here

packages/testkit/src/ckb-indexer-helper.ts
```diff
@@ -0,0 +1,38 @@
+import shell from "shelljs";
+import compareVersions from "compare-versions";
+
+let CKB_Indexer_Version = "0.2.2";
```
homura, line 4: we should avoid global variable here

homura: optimize it in next PR

PR #280, opened 2022-03-03
fix: transaction collector input when same cell in != transactions

jpeersyst:
I have talked about this PR in this issue: [Transaction collector bugfix](https://github.com/nervosnetwork/lumos/issues/278)
This is the implementation of what I commented there

### DIFF
--- a/packages/ckb-indexer/src/transaction_collector.ts
+++ b/packages/ckb-indexer/src/transaction_collector.ts
@@ -103,19 +103,23 @@ export class CKBIndexerTransactionCollector extends BaseIndexerModule.Transactio
       transactionHashList
     );
     //get input cell transaction batch
-    const txIoTypeInputOutPointList: unknown[] = [];
+    const txIoTypeInputOutPointList: any[] = [];
     transactionList.forEach((transactionWrapper) => {
       if (transactionWrapper.ioType === "input") {
         const targetOutPoint: OutPoint =
           transactionWrapper.transaction.inputs[
             parseInt(transactionWrapper.ioIndex)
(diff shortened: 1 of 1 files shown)


packages/ckb-indexer/src/transaction_collector.ts
```diff
@@ -103,19 +103,23 @@ export class CKBIndexerTransactionCollector extends BaseIndexerModule.Transactio
...
           ].previous_output;
-        txIoTypeInputOutPointList.push({
-          id: targetOutPoint.index + "-" + transactionWrapper.transaction.hash,
-          jsonrpc: "2.0",
-          method: "get_transaction",
-          params: [targetOutPoint.tx_hash],
-        });
+        const id =
+          targetOutPoint.index + "-" + transactionWrapper.transaction.hash;
+        if (!txIoTypeInputOutPointList.some((txReq) => txReq.id === id)) {
```
sighWang, line 115: id just include transaction_hash and outpint.Index, not include transaction_ioIndex, so we maybe lose some request which transaction is same but transaction.ioIndex not.

I have created a new PR https://github.com/nervosnetwork/lumos/pull/286 to fix this problem. 
And also I have created https://github.com/nervosnetwork/lumos/pull/282 to add some test.

jpeersyst: Thanks for the fast reply and fix! I have just seen the PR

jpeersyst: I am closing this PR as it is has been solved in this other two

PR #282, opened 2022-03-04
fix(ckb-indexer): TxCollector cannot resolve OutPoint

sighWang:
when more than one input needs to be resolved, the current TransactionCollector will only resolve the first one, which causes an error in the program

according to this bug report issue https://github.com/nervosnetwork/lumos/issues/278 and bug fix PR https://github.com/nervosnetwork/lumos/pull/280

### DIFF
--- a/packages/ckb-indexer/tests/transaction_collector.test.ts
+++ b/packages/ckb-indexer/tests/transaction_collector.test.ts
@@ -198,7 +198,7 @@ test("input cell can be found in transaction detail", async (t) => {
 });
 
 test("should add inputCell to all transaction which txHash and ioType is same as query", async (t) => {
-  const args5 = {
+  const searchKey = {
     script: {
       code_hash:
         "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
@@ -208,8 +208,10 @@ test("should add inputCell to all transaction which txHash and ioType is same as
     script_type: "lock" as ScriptType,
     filter: {},
   };
-  const args6 = { sizeLimit: undefined, order: undefined };
-  getTransactionsStub.withArgs(args5, args6).returns(multipleInputCellTx);
+  const SearchFilter = { sizeLimit: undefined, order: undefined };
+  getTransactionsStub
+    .withArgs(searchKey, SearchFilter)
+    .returns(multipleInputCellTx);
   requestBatchStub
     .withArgs(nodeUri, batchForAll)
     .returns(multipleInputCellBatchRequestResult);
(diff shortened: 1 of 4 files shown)


packages/ckb-indexer/tests/transaction_collector.test.ts
```diff
@@ -137,19 +145,46 @@ test("throw error when pass wrong fromBlock(toBlock) to TransactionCollector", (
...
+  const searchKey1 = {
+    script: {
+      code_hash:
+        "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
+      hash_type: "type" as HashType,
+      args: "0xbde8b19b4505dd1d1310223edecea20adc4e240e",
+    },
+    script_type: "lock" as ScriptType,
+    filter: {},
+  };
```
homura:
```suggestion
  const searchKey1: SearchKey = {
    script: {
      code_hash:
        "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
      hash_type: "type",
      args: "0xbde8b19b4505dd1d1310223edecea20adc4e240e",
    },
    script_type: "lock",
    filter: {},
  };
```

homura: we should avoid using `as` to cast types as much as possible, although this is convenient, it is not a good practice

packages/ckb-indexer/tests/transaction_collector.test.ts
```diff
@@ -161,3 +196,43 @@ test("input cell can be found in transaction detail", async (t) => {
...
+      hash_type: "type",
+      args: "0xdfc8ad219178a307ff117ee4e2fe760cd18410a7",
+    },
+    script_type: "lock",
+    filter: {},
+  };
+  const SearchFilter = { sizeLimit: undefined, order: undefined };
+  getTransactionsStub
+    .withArgs(searchKey, SearchFilter)
+    .returns(multipleInputCellTx);
```
homura, line 214: maybe the returns should be wrapped with a Promise?

sighWang: yes,multipleInputCellTx is a Promise, it have be wrapped at https://github.com/sighWang/lumos/blob/fix-transaction-collector-bug/packages/ckb-indexer/tests/transaction_collector_special_test_case.ts#L278

homura: i see

jpeersyst:
Thanks for the fast solution guys!
If I see something else I'll try to do the test aswell

PR #286, opened 2022-03-07
fix(ckb-indexer): reslove correct cell to transaction

sighWang:


packages/ckb-indexer/src/transaction_collector.ts
```diff
@@ -113,31 +113,39 @@ export class CKBIndexerTransactionCollector extends BaseIndexerModule.Transactio
...
+            const [cellIndex, transactionHash, ioIndex] = itemId.split("-");
             const output: Output =
               item.result.transaction.outputs[parseInt(cellIndex)];
-            const targetTxs = transactionList.filter(
+            const targetTx = transactionList.find(
               (tx) =>
-                tx.transaction.hash === transactionHash && tx.ioType === "input"
+                tx.transaction.hash === transactionHash &&
+                tx.ioType === "input" &&
+                tx.ioIndex === ioIndex
```
homura:
```suggestion
                Number(tx.ioIndex) === Number(ioIndex)
```

packages/ckb-indexer/src/transaction_collector.ts
```diff
@@ -113,31 +113,39 @@ export class CKBIndexerTransactionCollector extends BaseIndexerModule.Transactio
...
+          method: "get_transaction",
+          params: [targetOutPoint.tx_hash],
+        });
       }
     });
     if (txIoTypeInputOutPointList.length > 0) {
       await services
         .requestBatch(this.CKBRpcUrl, txIoTypeInputOutPointList)
         .then((response: GetTransactionRPCResult[]) => {
+          console.log(response.length);
```
homura:
```suggestion
```

packages/ckb-indexer/tests/transaction_collector.test.ts
```diff
@@ -190,7 +190,7 @@ test("input cell can be found in transaction detail", async (t) => {
     .returns(ioTypeInputResult);
   const cellCollector = new TransactionCollector(indexer, queryOption, nodeUri);
   const count = await cellCollector.count();
-  t.is(count, 2);
+  t.is(count, 1);
```
sighWang, line 193: change `2` to `1 `cause I delete some test data, 
now the test flow change to I have only one transaction that` ioType` is `input`. and finally I count the result, there is also `1`.
maybe I can change `1` to a variable named `expectedCount`

homura: why we should DELETE test data?

sighWang: this test aimed to test how we resolved transaction if we have indexerTransaction which  ioType is input, so one indexerTransaction which ioType is input is enough.

I have create a new PR to write easier to understand unit test https://github.com/sighWang/lumos/pull/41

PR #297, opened 2022-03-11
fix: export deploy generator

liubq7:


packages/common-scripts/src/index.ts
```diff
@@ -8,6 +8,12 @@ import sudt from "./sudt";
 import anyoneCanPay from "./anyone_can_pay";
 import { parseFromInfo } from "./from_info";
 import { createP2PKHMessageGroup } from "./p2pkh";
+import {
+  generateDeployWithDataTx,
+  generateDeployWithTypeIdTx,
+  generateUpgradeTypeIdDataTx,
+  compareScriptBinaryWithOnChainData,
+} from "./deploy";
```
homura: we should keep the `export` style consistent

```ts
import deploy from './deploy'

...

export {
  ...,
  deploy
}
```

PR #318, opened 2022-03-28
perf(ckb-indexer): filter out unnecessary indexer data requests

sighWang:
`@ckb-lumos/ckb-indexer` based on `ckb-indexer`,and ckb-indexer `outputDataLenRange` filed support  prefix match data len.
when you do `get_cells` action with data filed specify. this PR can you deduce  `outputDataLenRange` . so we can get_cell from indexer rpc more efficient.

for example. if you set your queries as below.
```js
{
    lock: {
        code_hash:
          "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
        hash_type: "type",
        args: "0xbde8b19b4505dd1d1310223edecea20adc4e240e",
      },
    data: "0x",
  }
```
lumos helps you to covert your queries  as below in background.

```js
{
    lock: {
        code_hash:
          "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
        hash_type: "type",
        args: "0xbde8b19b4505dd1d1310223edecea20adc4e240e",
      },
    data: "0x",
   outputDataLenRange: ["0x0" , "0x1"]
  }
```

### DIFF
--- a/packages/ckb-indexer/src/collector.ts
+++ b/packages/ckb-indexer/src/collector.ts
@@ -43,18 +43,91 @@ export class CKBCellCollector implements BaseCellCollector {
       bufferSize: undefined,
     };
     this.queries = { ...defaultQuery, ...this.queries };
-    if (
-      !this.queries.lock &&
-      (!this.queries.type || this.queries.type === "empty")
-    ) {
+    this.validateParams(this.queries);
+    this.convertParams();
+  }
+
+  public validateParams(queries: CKBIndexerQueryOptions): void {
+    if (!queries.lock && (!queries.type || queries.type === "empty")) {
       throw new Error("Either lock or type script must be provided!");
     }
 
+    if (queries.lock) {
+      if (!instanceOfScriptWrapper(queries.lock)) {
+        validators.ValidateScript(queries.lock);
+      } else if (instanceOfScriptWrapper(queries.lock)) {
+        validators.ValidateScript(queries.lock.script);
+      }
+    }
+
+    if (queries.type && queries.type !== "empty") {
+      if (
+        typeof queries.type === "object" &&
+        !instanceOfScriptWrapper(queries.type)
+      ) {
+        validators.ValidateScript(queries.type);
+      } else if (
+        typeof queries.type === "object" &&
+        instanceOfScriptWrapper(queries.type)
+      ) {
+        validators.ValidateScript(queries.type.script);
+      }
+    }
+
+    if (queries.fromBlock) {
+      utils.assertHexadecimal("fromBlock", queries.fromBlock);
+    }
+    if (queries.toBlock) {
+      utils.assertHexadecimal("toBlock", queries.toBlock);
+    }
+    if (queries.order !== "asc" && queries.order !== "desc") {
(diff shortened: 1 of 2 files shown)


packages/ckb-indexer/src/collector.ts
```diff
@@ -43,82 +43,117 @@ export class CKBCellCollector implements BaseCellCollector {
       bufferSize: undefined,
     };
     this.queries = { ...defaultQuery, ...this.queries };
-    if (
-      !this.queries.lock &&
-      (!this.queries.type || this.queries.type === "empty")
-    ) {
+    this.validateParams(this.queries);
+    this.convertParams();
```
```suggestion
    this.validateQueryOption(this.queries);
    this.convertQueryOptionToSearchKey();
```

The method names here can be confusing, in the implementation of `@ckb-lumos/ckb-indexer` we should distinguish between `SearchKey`(for indexer) and `QueryOption`(for client)

packages/ckb-indexer/src/collector.ts
```diff
@@ -43,82 +43,117 @@ export class CKBCellCollector implements BaseCellCollector {
...
+
+    if (!this.queries.outputDataLenRange) {
+      if (this.queries.data && this.queries.data !== "any") {
+        const dataLenRange = getHexStringBytes(this.queries.data);
+        this.queries.outputDataLenRange = [
+          "0x" + dataLenRange.toString(16),
+          "0x" + (dataLenRange + 1).toString(16),
+        ];
+      }
+    }
```
homura, line 154: we should also check if both `data` and `outputDataLenRange` are present in the `QueryOption`

```ts
// pseudocode
if (outputDataLenRange && data) {
  asserts(
    byteLengthOf(data) >= outputDataLenRange[0] &&
      byteLengthOf(data) < outputDataLenRange[1],
    throwError("mismatched datalen and range ...")
  );
}
```

sighWang: validate logic has been done in the [validation function ](https://github.com/nervosnetwork/lumos/pull/318/files#diff-b6c230b5ef87aecafc6364827f77b3983aa96df957e497617063ca55a072fe92R108) 

and I add test at [next commit](https://github.com/nervosnetwork/lumos/pull/318/commits/21f80dc7535ca0277e741479e423d793dbd32168)

packages/ckb-indexer/tests/cell_collector.unit.test.ts
```diff
@@ -6,15 +6,16 @@ const nodeUri = "http://127.0.0.1:8118/rpc";
...
 const indexer = new Indexer(indexUri, nodeUri);
 
+const lockScript: Script = {
+  code_hash:
+    "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
+  hash_type: "type",
+  args: "0xbde8b19b4505dd1d1310223edecea20adc4e240e",
+};
+
 test.serial(
```
homura: why use `test.serial("...")` instead of `test("...")`？

PR #326, opened 2022-04-02
fix: hd-cache constract failed

sighWang:
BREAKING CHANGE: class `Cache `and `CacheManager `from `@ckb-lumos/hd-cache `  changed method params

detail change logs are blowing,and migration docs can be find at[ readme page](https://github.com/nervosnetwork/lumos/tree/develop/packages/hd-cache#usage)
### Cache constructor
change optional TransactionCollector Class to required and add type `typeof BaseTransactionCollector`
```diff
import {
  HexString,
  Indexer,
  TransactionCollector as BaseTransactionCollector
} from "@ckb-lumos/base";
constructor(
    indexer: Indexer,
    publicKey: HexString,
    chainCode: HexString,
    infos: LockScriptMappingInfo[],
    {
-     TransactionCollector,
      masterPublicKey,
      rpc,
    }: {
-     TransactionCollector: any;
+    transactionCollector: typeof BaseTransactionCollector;
      masterPublicKey?: HexString;
      rpc?: RPC;
    }
  );
```

### CacheManager constructor
change optional TransactionCollector Class to required and add type `typeof BaseTransactionCollector`, move `masterPublicKey ` to end.
```diff
constructor(
    indexer: Indexer,
    publicKey: HexString,
    chainCode: HexString,
-  masterPublicKey?: HexString,
    infos:  LockScriptMappingInfo[] | undefined,
    {
      logger,
      pollIntervalSeconds,
      livenessCheckIntervalSeconds,
      TransactionCollector,
      rpc,
    }?: {
      logger?: (level: string, message: string) => void;
      pollIntervalSeconds?: number;
      livenessCheckIntervalSeconds?: number;
-    TransactionCollector?: any;
+    TransactionCollector: typeof BaseTransactionCollector;
      rpc?: RPC;
    },
+   masterPublicKey?: HexString
  );
```
### CacheManager.loadFromKeystore and CacheManager.fromMnemonic

change optional TransactionCollector Class to required and add type `typeof BaseTransactionCollector`,

```diff
static loadFromKeystore(
    indexer: Indexer,
    path: string,
    password: string,
    infos: LockScriptMappingInfo[] | undefined,
    options: {
      logger?: (level: string, message: string) => void;
      pollIntervalSeconds?: number;
      livenessCheckIntervalSeconds?: number;
-    TransactionCollector?: any;
+     TransactionCollector: typeof BaseTransactionCollector;
      needMasterPublicKey?: boolean;
      rpc?: RPC;
    }
  ): CacheManager;

  static fromMnemonic(
    indexer: Indexer,
    mnemonic: string,
    infos: LockScriptMappingInfo[] | undefined,
    options: {
      logger?: (level: string, message: string) => void;
      pollIntervalSeconds?: number;
      livenessCheckIntervalSeconds?: number;
-   TransactionCollector?: any;
+   TransactionCollector: typeof BaseTransactionCollector;
      needMasterPublicKey?: boolean;
      rpc?: RPC;
    }
  ): CacheManager;
```


packages/common-scripts/src/anyone_can_pay.ts
```diff
@@ -32,11 +32,13 @@ import {
...
+import { CellCollectorConstructor, CellCollectorType } from "./type";
 const { ScriptValue } = values;
 const { CKBHasher, ckbHash, readBigUInt128LECompatible } = utils;
 
-export class CellCollector implements CellCollectorType {
-  private cellCollector: CellCollectorType;
+export const CellCollector: CellCollectorConstructor = class CellCollector
+  implements CellCollectorType {
+  private cellCollector: BaseCellCollectorType;
   private config: Config;
```
homura, line 42: maybe we can implement more than one interface

```ts
export class CellCollector implements CellCollectorType, CellCollectorConstructor {}
```

packages/common-scripts/src/type.ts
```diff
@@ -0,0 +1,27 @@
...
+    }: Options & {
+      queryOptions?: QueryOptions;
+    }
+  ): CellCollectorType;
+}
+
+export interface CellCollectorType extends CellCollector {
+  readonly fromScript: Script;
+  collect(): AsyncGenerator<Cell>;
+}
```
homura: why are constructor and instance defined separately?

packages/common-scripts/src/anyone_can_pay.ts
sighWang: we can not  implements `CellCollectorConstructor` cause it's [static side](https://www.typescriptlang.org/docs/handbook/interfaces.html#difference-between-the-static-and-instance-sides-of-classes).  and also we need `CellCollector`  from '`anyone_one_can_pay`' ,'`secp256k1_blake160_multisig`' and '`secp256k1_blake160`' as [same type](https://github.com/sighWang/lumos/blob/fix-lint-error/packages/common-scripts/src/common.ts#L48). so we write `CellCollector: CellCollectorConstructor` here

packages/common-scripts/src/type.ts
sighWang: `CellCollectorConstructor` for class type as` CellCollector: CellCollectorConstructor`,and `CellCollectorType` for instance side.
maybe I should change `CellCollectorType` to `CellCollectorInterface`

sighWang: I have simplify the problem as blow

<details>
<summary> how to define type of Super.A  </summary>

```ts
interface A {
  a(): void;
}

export class A1 implements A {
  private name: string;
  age: number;
  constructor(name: string) {
    this.name = name;
    this.age = 19;
  }
  a(): void {
    console.log("A1");
  }
}

export class A2 implements A {
  private name: string;
  constructor(name: string) {
    this.name = name;
  }
  a(): void {
    console.log("A2");
  }
}

export class A3 implements A {
  private name: string;
  constructor(name: string) {
    this.name = name;
  }
  a(): void {
    console.log("A3");
  }
}

interface Super {
  A: any; // how to define type of A,so we can assign A1,A2,A3 to A
}
class Test {
  super: Super = { A: A2 };
  test() {
    this.super.A = A2;
    this.super.A = A3;
    this.super.A = A1;
  }
}
```
</details>

packages/rpc/src/index.ts
```diff
@@ -87,7 +89,7 @@ class RpcProxy {
     }
     const header: Header = await this.rpc.get_tip_header();
     const blockNumber = BI.from(header.number);
-    while (true) {
+    for (;;) {
```
zhangyouxin: why refactor here? some conventions?

packages/hd-cache/tests/cache.test.ts
```diff
@@ -186,7 +177,11 @@ const mnemonic =
  * 3: 0x57a81755c7229decb0f21f93d73c1c7e1c0afe95
  */
 
-class MockTransactionCollector extends TransactionCollector {
+const nodeUri = "htp://127.0.0.1:8118/rpc";
+const indexUri = "ttp://127.0.0.1:8120";
+const indexer = new CkbIndexer(indexUri, nodeUri);
```
zhangyouxin: Maybe use RPC_URL and INDEXER_URL as viarable name here

packages/rpc/src/index.ts
sighWang: `while(true)` will trigger eslint error 'no-constant-condition'

zhangyouxin: Okay, looks reasonable, but I would say that `while (true)` is more intuitive.

Can we change the rule in `.eslintrc.js` to `no-constant-condition: ["error", { "checkLoops": false }]` here?

Reference: https://eslint.org/docs/rules/no-constant-condition#checkloops

packages/common-scripts/tests/common.test.ts
```diff
@@ -32,6 +34,29 @@ test.before(() => {
     throw new Error("can not find bigint");
   };
 });
+let callCount = 0;
+sinon.stub(CellCollector.prototype, "collect").callsFake(async function* () {
+  callCount++;
+  if (callCount === 1) {
```
homura: i think this PR is just a lint fix, why tests must change here?

packages/hd-cache/tests/cache.test.ts
```diff
@@ -208,10 +203,19 @@ class MockTransactionCollector extends TransactionCollector {
   }
 }
 
-const indexer = new MockIndexer();
-
+stub(RPC.prototype, "get_header").callsFake(async function (
```
homura, line 206: i think we should AVOID using `stub` at all costs, and only use it if the mock can't fulfill the requirement, such as when the code depend on global something like `globalThis.XMLHttpRequest`, otherwise it's a means of testing for the sake of testing, not testing for correctness

packages/common-scripts/tests/common.test.ts
sighWang: use `sinon stub` and `callCount `is wrong here,I change bank to use `MockCellCollector`

packages/hd-cache/tests/cache.test.ts
sighWang: I think `stub `here is fine, it's just return headers. If we use `mock`, we will do what `stub `do as well. I have some wrong usage of `stub `like `callCount+stub`, And I have remove them as you suggested

packages/common-scripts/src/common.ts
```diff
@@ -9,7 +9,9 @@ import secp256k1Blake160Multisig from "./secp256k1_blake160_multisig";
 import { FromInfo, parseFromInfo } from "./from_info";
 import secp256k1Blake160 from "./secp256k1_blake160";
 import { getConfig, Config } from "@ckb-lumos/config-manager";
-import locktimePool from "./locktime_pool";
+import locktimePool, {
+  CellCollector as locktimePoolCellCollector,
```
homura:
```suggestion
  CellCollector as LocktimePoolCellCollector,
```

packages/hd-cache/README.md
```diff
@@ -34,3 +34,27 @@ for await (const cell of cellCollector.collect()) {
...
 await getBalance(cellCollector)
 ```
+
+### how to quickly get transactionCollector
+
+`CKBIndexerTransactionCollector.asBaseTransactionCollector` helps you to get a transactionCollector instance.
+
+```ts
+const nodeUri = "htp://127.0.0.1:8118/rpc";
+const indexUri = "ttp://127.0.0.1:8120";
```
homura:
```suggestion
const CKB_URL = "http://127.0.0.1:8118/rpc";
const CKB_INDEXER_URL = "http://127.0.0.1:8120";
```

packages/hd-cache/src/index.ts
```diff
@@ -201,16 +202,24 @@ export class HDCache {
...
   getNextReceivingPublicKeyInfo(): PublicKeyInfo {
-    const info: PublicKeyInfo = this.receivingKeys.find(
+    const info: PublicKeyInfo | undefined = this.receivingKeys.find(
       (key) => key.historyTxCount === 0
-    )!;
+    );
+    /* c8 ignore next 3 */
+    if (!info) {
+      throw new Error("Impossible: can not find receivingKey");
+    }
```
homura: i find there are many of if(!condition) { throw new impossible error }, would you please refactor to assertion style like this

```ts
function assertsNonNil(value: unknown): asserts value is NonNullable<T> {
  if (value === undefined || value === null) throw new Error("...")
}
```

packages/common-scripts/src/common.ts
sighWang: already have `LocktimePoolCellCollector`  in file, Can I change to `LocktimePoolCellCollectorClass`?

homura: maybe we can change it to `LocktimePollCellCollectorConstructor` ?

PR #352, opened 2022-06-15
feat: implement omnilock p2pkh

zhangyouxin:
implement p2pkh adjust txSkelenton


packages/omnilock/package.json
```diff
@@ -21,7 +21,8 @@
   "dependencies": {
     "@ckb-lumos/base": "*",
     "@ckb-lumos/codec": "*",
-    "@ckb-lumos/common-scripts": "*"
+    "@ckb-lumos/common-scripts": "*",
+    "ethers": "^5.6.9"
```
homura, line 25: why we need ethers here?

packages/omnilock/examples/omnilock-metamask.ts
```diff
@@ -50,7 +50,7 @@ async function main() {
     },
   });
 
-  let txSkeleton = TransactionSkeleton({});
+  let txSkeleton = new TransactionSkeleton({});
```
homura, line 53: maybe the `new` is incorrect

packages/omnilock/src/suite/index.ts
```diff
@@ -57,29 +54,62 @@ export function createDefaultOmnilockSuite(
     },
 
     async adjust(txSkeleton) {
-      let adjustedSkeleton = p2pkh(txSkeleton, {
+      const config = getConfig();
```
homura, line 57: we should move the `config` to the parameter because `getConfig` is actually a global variable, this is a historical problem that makes it difficult for users to troubleshoot if they accidentally use multiple versions of `config-manager` in their projects

PR #353, opened 2022-06-15
feat(examples): add omnilock administrator mode example

PainterPuppets:
- [x] add omnilock administrator mode example
- [x] add example doc

### DIFF
--- a/examples/omni-lock-administrator/index.ts
+++ b/examples/omni-lock-administrator/index.ts
@@ -1,0 +1,369 @@
+import {
+  config,
+  Script,
+  Address,
+  helpers,
+  Indexer,
+  RPC,
+  commons,
+  hd,
+  toolkit,
+  HexString,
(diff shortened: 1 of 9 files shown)


examples/omni-lock-administrator/index.ts
```diff
@@ -0,0 +1,329 @@
...
+  return {
+    lockScript,
+    address,
+    pubKey,
+    privKey,
+  };
+};
+
+export const alice = generateSECP256K1Account("0xfd686a48908e8caf97723578bf85f746e1e1d8956cb132f6a2e92e7234a2a245");
+export const bob = generateSECP256K1Account("0xfd686a48908e8caf97723578bf85f746e1e1d8956cb132f6a2e92e7234a2a245");
```
homura: maybe we can use a different key to distinguish Alice from Bob

PR #357, opened 2022-06-21
feat(debugger): supported dep_group in TestContext

homura:
## Motivation 

We usually split a large contract into several subcontracts for deployment, and depend on them as [dep_group](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0022-transaction-structure/0022-transaction-structure.md#dep-group)

## Example

```ts
const { executor } = createTestContext({
  deps: {
    SECP256K1_BLAKE160: {
      dep_type: "dep_group",
      path: path.join(__dirname, "deps/secp256k1_blake160"),
      includes: [path.join(__dirname, "deps/secp256k1_data_info")],
    },
  }
})

txSkeleton = txSkeleton.update('inputs', inputs => inputs.push( /* secp256k1 lock */)

// ...

executor.execute(txSkeleton)
```

for more details, we can check this [case](https://github.com/nervosnetwork/lumos/compare/debugger-dep-group?expand=1#diff-c9daacd74eadd3a64c8bb79570771d6cc0b4fa6795c66327fe8c3688827e2180R116-R151)


PainterPuppets (approved):
LGTM!

zhangyouxin (approved):
LGTM

PR #358, opened 2022-06-22
feat(debugger): supported debug message

homura:
## Motivation

When we execute a transaction, we may find that the return code is an error code, but we do not know the exact cause of the error. In this case, we may [compile the contract with debug message](https://github.com/nervosnetwork/ckb-vm/discussions/193) and then debug it.

This PR allows us to get the debug message from the `ExecuteResult`

## Example

```js
const { debugMessage } = await executor.execute(txSkeleton);
console.log(debugMessage)
```

for more details, we can check out [this example](https://github.com/nervosnetwork/lumos/pull/358/commits/1a1760f4aa9f376cfc772e89382de5c99f8b59f7#diff-c9daacd74eadd3a64c8bb79570771d6cc0b4fa6795c66327fe8c3688827e2180R199-R220), which shows how to output the debug message in [debugger.c](https://github.com/nervosnetwork/ckb/blob/900a7bf6447443a465b15039ef12de8bf73fee33/script/testdata/debugger.c)


zhangyouxin (approved):
LGTM

PR #361, opened 2022-07-05
feat: add debugger default config

zhangyouxin:
motivation: users need to copy contract binaries and make their own `config` if they want to `createTestContext(config)`, by adding a default config,  creating a common TestContext would be more convenient.

This PR adds:

- ALWAYS_SUCCESS
- ALWAYS_FAIL
- ANYONE_CAN_PAY
- SECP256K1_BLAKE160
- SECP256K1_BLAKE160_MULTISIG
- DAO
- SUDT
- OMNI_LOCK

to contract list in default config.

---

Added test case for:

- secp256k1 multisig


packages/debugger/src/context.ts
```diff
@@ -130,3 +131,52 @@ export function createTestContext<Code extends LocaleCode>(config: {
...
+      },
+      ANYONE_CAN_PAY: {
+        dep_type: "code",
+        path: path.join(__dirname, "../bin/anyone_can_pay"),
+      },
+      SUDT: {
+        dep_type: "code",
+        path: path.join(__dirname, "../bin/sudt"),
+      },
+      OMNI_LOCK: {
```
homura:
```suggestion
      OMNILOCK: {
```
the Omnilock is named “Omnilock” in RFC, so we can just use `OMNILOCK` here

packages/debugger/src/context.ts
```diff
@@ -130,3 +131,52 @@ export function createTestContext<Code extends LocaleCode>(config: {
...
 }
+
+export function getDefaultConfig(): {
+  deps: LocaleCode;
+} {
+  return {
+    deps: {
+      ALWAYS_SUCCESS: {
+        dep_type: "code",
+        path: path.join(__dirname, "../bin/always_success"),
```
homura: the `context.ts` will be transpiled to `lib/context.js`, we should add `bin` to `package.json`'s `files` to ensure `bin` will be included when published

homura (approved):

PR #365, opened 2022-07-18
feat(codec): molecule unpack works with BytesLike

homura:
## Motivation

When we pack or unpack, the molecule codec only allows us to pass in qualified types like `HexString` or `Uint8Array` to represent binary data.
The codec would be more convenient if we could pass `BytesLike` directly

## How

```ts
const RGB = struct({ r: Uint8, g: Uint8, b: Uint8 }, ["r", "g", "b"]);

// the array of number is OK
console.log(RGB.unpack([255, 255, 255]));

// also the HexString is OK
console.log(RGB.unpack("0xffffff"));
```

```ts
// with 65 bytes lock placehoder
WitnessArgs.pack({
  lock: new Uint8Array(65)
})
```

### DIFF
--- a/packages/codec/src/base.ts
+++ b/packages/codec/src/base.ts
@@ -49,7 +49,7 @@ export type UnpackParam<T extends AnyCodec> = T extends Codec<
   ? Unpackable
   : never;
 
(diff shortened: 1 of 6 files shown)


zhangyouxin (approved):
LGTM

PR #372, opened 2022-07-25
feat: add missing function of merge ckb-sdk-utils

PainterPuppets:


packages/helpers/src/index.ts
```diff
@@ -28,6 +28,31 @@ export interface Options {
 
 const BECH32_LIMIT = 1023;
 
+export function minimalScriptCapacity(
+  script: Script,
+  { validate = true }: { validate?: boolean } = {}
+): bigint {
```
zhangyouxin, line 36: Using bigint is a legacy issue, @homura maybe we should change all bigint to BI? so that we don't need to add xxxCapatible in the future.

homura:
would you please fix the blocking CI by Vercel?

packages/helpers/src/index.ts
homura: we'll keep the API same as the other functions for now, so we won't change it for now, let's keep `minimalScriptCapacity` + `minimalScriptCapacityCompatible`

homura:
@PainterPuppets can you add some unit test for these new features to make CI pass

PainterPuppets:
change merge base to branch `0.19`

packages/common-scripts/src/deploy.ts
```diff
@@ -359,7 +359,7 @@ interface ScriptConfig {
   SHORT_ID?: number;
 }
 
-function calculateTxHash(txSkeleton: TransactionSkeletonType): string {
+export function calculateTxHash(txSkeleton: TransactionSkeletonType): string {
```
zhangyouxin, line 362: why we should expose this utility function in `deploy.ts` here?

packages/common-scripts/src/dao.ts
```diff
@@ -602,7 +602,7 @@ function _addDaoCellDep(
   });
 }
 
-function extractDaoDataCompatible(dao: PackedDao): {
+export function extractDaoDataCompatible(dao: PackedDao): {
```
zhangyouxin, line 605: Do we really need to expose this function in `dao.ts` here?

packages/helpers/tests/minimal_cell_capacity.test.ts
```diff
@@ -83,3 +88,36 @@ test("cell with type and data, validate true", (t) => {
...
+      "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
+    hashType: "type",
+  });
+  const expectedCapacity = BI.from(53).mul(BI.from(10).pow(8));
+  t.true(capacity.toString() === expectedCapacity.toString());
+});
+
+test("invalid script, validate failed", (t) => {
+  t.throws(() => {
+    minimalScriptCapacity({
```
zhangyouxin: If we decide to test `minimalScriptCapacity` in  `packages/helpers/tests/minimal_cell_capacity-bigint.test.ts`, then this test case should be moved to that file.

PainterPuppets: LGTM

zhangyouxin (approved):
LGTM

PR #378, opened 2022-08-04
refactor!: camalize exposed APIs

zhangyouxin:
This pr has:
- Refactor user interfaces to camelCase
  - Changed types in `api.d.ts` in `@ckb-lumos/base` to camelCase
  - Changed types used in `@ckb-lumos/ckb-indexer` to camelCase
  - Added param formatter and result formatter in `@ckb-lumos/ckb-indexer` to support camelCase
- Removed old serialization method
  - Removed `core.js` in `@ckb-lumos/base`
  - Removed `Normalizers` and `Transformers` in `@ckb-lumos/toolkit`
  - Implemented new serialization methods in `@ckb-lumos/base/blockchain.ts`
- Replaced `@ckb-lumos/rpc` with RPC module in https://github.com/nervosnetwork/ckb-sdk-js


.gitignore
```diff
@@ -116,6 +116,7 @@ packages/experiment-tx-assembler/lib
 
 # knex config
 packages/sql-indexer/knexfile.ts
+packages/lib
```
homura: maybe it is unnecessary

packages/base/src/utils.js
```diff
@@ -1,40 +1,32 @@
...
+    this.hasher = blake2b(32, null, null, bytifyRawString("ckb-default-hash"));
   }
 
   update(data) {
-    this.hasher.update(new Uint8Array(new Reader(data).toArrayBuffer()));
+    this.hasher.update(bytify(data));
     return this;
   }
 
-  digestReader() {
```
homura, line 23: we need to put this method into migration list

packages/base/src/utils.d.ts
```diff
@@ -1,18 +1,15 @@
...
-  update(data: string | Reader | ArrayBuffer): this;
-
-  digestReader(): Reader;
+  update(data: string | ArrayBuffer): this;
 
   digestHex(): Hash;
 }
 
-export function ckbHash(buffer: ArrayBuffer): Reader;
+export function ckbHash(buffer: ArrayBuffer): Hash;
```
homura: we need to put this method into migration list

packages/base/tests/blockchain.test.ts
```diff
@@ -83,3 +86,64 @@ test("a real world Omni Lock witness should work as expected", (t) => {
   t.is(omniLockWitnessPlaceholder.byteLength, 105);
   t.deepEqual(packedWitness, omniLockWitnessPlaceholder);
 });
+
+test("Script codec", (t) => {
```
homura: these cases are same duplicated with serialization.test.ts, maybe we can remove them

packages/ckb-indexer/src/collector.ts
```diff
@@ -125,11 +125,11 @@ export class CKBCellCollector implements BaseCellCollector {
...
   public convertQueryOptionToSearchKey(): void {
+    const queryLock = this.queries.lock;
     // unWrap `ScriptWrapper` into `Script`.
-    if (this.queries.lock) {
-      if (instanceOfScriptWrapper(this.queries.lock)) {
-        validators.ValidateScript(this.queries.lock.script);
-        this.queries.lock = this.queries.lock.script;
+    if (queryLock) {
+      if (instanceOfScriptWrapper(queryLock)) {
+        this.queries.lock = queryLock.script;
```
homura, line 133:
```suggestion
        validators.ValidateScript(queryLock.script);
        this.queries.lock = queryLock.script;
```

packages/ckb-indexer/src/indexer.ts
```diff
@@ -241,7 +244,7 @@ export class CkbIndexer implements Indexer {
       });
   }
 
-  private scheduleLoop(timeout = 1) {
+  scheduleLoop(timeout = 1): void {
```
homura: any reason for removing the `private` symbol modifier here?

packages/ckb-indexer/tests/indexer.unit.test.ts
```diff
@@ -0,0 +1,54 @@
...
+      codeHash:
+        "0x0000000000000000000000000000000000000000000000000000000000000001",
+      hashType: "type" as HashType,
+      args: "0x0000000000000000000000000000000000000000000000000000000000000002",
+    },
+  };
+  const result: IndexerEmitter = indexer.subscribe(queryOption);
+
+  t.deepEqual(result.lock, queryOption.lock);
+  t.deepEqual(result.type, undefined);
```
homura, line 27: why we should care about the return of `indexer.subscribe(queryOption)` rather than `result.on(...)`

packages/ckb-indexer/tests/transaction_collector.e2e.test.ts
```diff
@@ -84,7 +84,11 @@ test("query transactions with different queryOptions", async (t) => {
     for await (const hash of transactionCollector.collect()) {
       transactionList.push(hash);
     }
-    t.deepEqual(transactionList, queryCase.expectedResult, queryCase.desc);
+    t.deepEqual(
+      utils.deepCamelizeTransaction(utils.deepCamel(transactionList)),
```
homura: Can we implement the `deepCamelizeTransaction(transactionList)` rather than `utils.deepCamelizeTransaction(utils.deepCamel(transactionList))`. The `deepCamel` for `deepCamel` object keys only, and the `deepCamelizeTransaction` for both key and value like `depGroup`

packages/common-scripts/src/deploy.ts
```diff
@@ -384,9 +370,7 @@ function getScriptConfigByDataHash(
   outputIndex: number
 ): ScriptConfig {
   const data = txSkeleton.outputs.get(outputIndex)!.data;
-  const codeHash = utils
-    .ckbHash(new Reader(data).toArrayBuffer())
-    .serializeJson();
+  const codeHash = utils.ckbHash(bytes.bytify(data));
```
homura, line 373: maybe we can make the `ckbHash` to support `BytesLike`

packages/debugger/src/types.ts
```diff
@@ -1,17 +1,9 @@
...
-  HexString,
-  Input,
-  OutPoint,
-  Output,
-  Transaction,
-} from "@ckb-lumos/base";
+import { Hash, Header, HexString, OutPoint } from "@ckb-lumos/base";
 import { CKBDebugger } from "./executor";
 import { LocaleCode } from "./context";
+import { RPC } from "@ckb-lumos/rpc/lib/types/rpc";
```
homura:
```suggestion
import type { RPC } from "@ckb-lumos/rpc/lib/types/rpc";
```

packages/ckb-indexer/src/indexer.ts
zhangyouxin: need to be stubbed so that I can write a test for `subscribe` method

homura: i see, but maybe we can add a `ts-ignore` for stub cases. we should only expose APIs that are useful to developers to avoid confusion for developers

packages/ckb-indexer/tests/indexer.unit.test.ts
zhangyouxin: The `result` didn't emit any events in `subscribe` method

PR #388, opened 2022-08-25
refactor(examples): secp256k1 transfer example

IronLu233:
# Description
The origin secp256k1 transfer example is too hard to understand. because it builds up the transaction mannually. In current lumos version, we have `commons.common.transfer` method, which can manipulate the raw transaction's inputs/outputs. Thus, it's a time for reactoring it. And I add a new feature, which allows developer can transfer CKB to mulitple addresses
# Type of change
[x] other


examples/secp256k1-transfer/src/lib.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/**
+ * sign a txSkeleton and send it to test network
+ * @returns Promise with transaction hash
+ */
+export async function transfer(txSkeleton: helpers.TransactionSkeletonType, privateKey: string): Promise<string> {
+  txSkeleton = commons.common.prepareSigningEntries(txSkeleton);
+  const message = txSkeleton.get("signingEntries").get(0)?.message;
+  const Sig = hd.key.signRecoverable(message!, privateKey);
```
homura:
```suggestion
  const sig = hd.key.signRecoverable(message!, privateKey);
```

examples/secp256k1-transfer/package.json
```diff
@@ -5,21 +5,32 @@
...
   },
   "devDependencies": {
     "@babel/cli": "^7.16.0",
     "@babel/core": "^7.16.0",
-    "@babel/plugin-transform-modules-commonjs": "^7.16.0"
+    "@babel/plugin-transform-modules-commonjs": "^7.16.0",
+    "parcel": "^2.7.0",
+    "postcss": "^8.4.16",
+    "process": "^0.11.10",
+    "typescript": "^4.7.4"
```
homura: can we remove these deps? maybe we can use the `examples/package.json` to keep the example's env same with others

examples/secp256k1-transfer/package.json
```diff
@@ -5,21 +5,32 @@
...
     "lint": "tsc --noEmit"
   },
   "keywords": [],
   "author": "",
   "license": "MIT",
   "dependencies": {
     "@ckb-lumos/lumos": "0.18.0",
+    "@headlessui/react": "^1.6.6",
+    "@heroicons/react": "^1.0.6",
+    "@mui/material": "^5.10.1",
```
homura: maybe these deps is unnecessary

examples/secp256k1-transfer/package.json
IronLu233: We can not remove it. When I remove typescript dependency, these code can not run in codesandbox.

homura: even so, please use the same version of TypeScript with [package.json](https://github.com/nervosnetwork/lumos/blob/635ca26c8b9daa5809a0afbe38b5a2ca462dda18/package.json#L43).
I found examples work in [codesandbox](https://codesandbox.io/s/github/nervosnetwork/lumos/tree/develop/examples/omni-lock-metamask?file=/package.json) now

IronLu233: Sorry, when I remove it, it can works well. The result of "Can not work" is from other reason.
https://codesandbox.io/s/github/IronLu233/lumos/tree/secp256k1-transfer/examples/secp256k1-transfer

examples/secp256k1-transfer/src/lib.ts
IronLu233: Rename it to `signature`. But why previous `Sig` name appears in master branch?

examples/secp256k1-transfer/src/lib.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  return balance;
+}
+
+interface Options {
+  targets: {
+    address: string;
+    amount: BIish;
+  }[];
+  privKey: string;
```
PainterPuppets: Here it should be address, not privkey, privkey -> address step shoud be place other

examples/secp256k1-transfer/src/lib.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  txSkeleton = await payFeeByFeeRate(txSkeleton, [transferFromInfo.address], 1000, undefined, { config: AGGRON4 });
+  return txSkeleton;
+}
+
+/**
+ * sign a txSkeleton and send it to test network
+ * @returns Promise with transaction hash
+ */
+export async function transfer(txSkeleton: helpers.TransactionSkeletonType, privateKey: string): Promise<string> {
```
PainterPuppets: The steps of signing and sending should be separated

examples/secp256k1-transfer/src/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+  createUnsignedTxSkeleton,
+  generateAccountFromPrivateKey,
+  transfer,
+  Account,
+  calculateTransactionFee,
+  MIN_CELL_CAPACITY,
+} from "./lib";
+
+type TransferTarget = {
+  amount: BI;
```
zhangyouxin: optional： capacity is more comprehensive than amount

examples/secp256k1-transfer/src/lib.ts
```diff
@@ -0,0 +1,119 @@
...
+    pubKey,
+  };
+};
+
+/**
+ * calculate the transaction fee by skeleton inputs and outputs
+ * @param skeleton unsigned transaction skeleton
+ * @returns the transaction fee, it is from `sum(skeleton inputs capacity) - sum(skeleton outputs capacity)`
+ */
+export function calculateTransactionFee(skeleton: helpers.TransactionSkeletonType) {
```
zhangyouxin, line 57: the logic in this function doesn't look like calculating tx fee to me.

zhangyouxin: `calculateTransactionFee` sounds like doing some calculation by tx size and fee rate, while the logic in this function is getting a paid tx fee, I suggest renaming it to `getPaidTxFee`

examples/secp256k1-transfer/src/lib.ts
```diff
@@ -0,0 +1,119 @@
...
+interface Options {
+  targets: {
+    address: string;
+    amount: BIish;
+  }[];
+  privKey: string;
+}
+
+/**
+ * create an unsigned 1:N transaction skeleton
```
zhangyouxin: not sure what is `1:N transaction skeleton` here

homura:
can you please upgrade the example to lumos@0.19? checkout the [migration guide](https://lumos-website.vercel.app/migrations/migrate-to-v0.19)

examples/secp256k1-transfer/src/lib.ts
```diff
@@ -0,0 +1,108 @@
...
+  const signature = hd.key.signRecoverable(message!, privateKey);
+  const tx = helpers.sealTransaction(txSkeleton, [signature]);
+  return tx;
+}
+
+/**
+ * send a transaction to CKB testnet
+ * @returns Promise with transaction hash
+ */
+export async function transfer(tx: Transaction): Promise<string> {
```
PainterPuppets: maybe this function is unnecessary, why not just use `rpc.send_transaction`?

IronLu233: I don't agree it. The three step: "create transaction", "sign",  and "send transaction" in one file can take more understanding to code reader.

examples/secp256k1-transfer/package.json
IronLu233: > ~~Sorry, when I remove it, it can works well. The result of "Can not work" is from other reason. https://codesandbox.io/s/github/IronLu233/lumos/tree/secp256k1-transfer/examples/secp256k1-transfer~~

These dependencies are needed when run it on local

examples/secp256k1-transfer/src/lib.ts
```diff
@@ -0,0 +1,108 @@
...
+  const hash = await rpc.sendTransaction(tx, "passthrough");
```
zhangyouxin: does this `rpc` need to be passed in from the function parameter?

IronLu233: I think it's no need. because the definition in https://github.com/nervosnetwork/lumos/pull/388/files#diff-23baadf37b78885a26cae62ec8dff08aa7dbd4083b9df75e84593d6adb404fa0R9 tells me it's a singleton.

homura:
IMO, as an example, it should be linear, procedural. Modular for engineering but not for viewing, especially for a web app like GitHub, which would require more memory if I had to jump between multiple methods.

Can you please refactor `transfer` into a method 

```ts
function transfer(options: {
  recipients: Array<{address: Address, capacity: BIish}>,
  senderPrivkey: BytesLike
}): Promise<Hash> {
  // step 1 - convert a privkey to an lock(address)

  // step 2 - caculate the total capacity to be transferred

  // step 3 - call common.transfer to generate a txSkeleton

  // step 4 - pay tx fee

  // step 5 - prepare signing entries

  // step 6 - seal the transaction

  // step 7 - broadcast the transfer tx
}


```

IronLu233:
> 

However, we need show the transaction fee. So we have to split the whole transfer into differient part.

examples/secp256k1-transfer/src/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+const createTransferTarget = (): TransferTarget => ({ key: nanoid(), capacity: MIN_CELL_CAPACITY, address: "" });
+
+export function App() {
+  const [state, setState] = useSetState({
+    privateKey: "",
+    accountInfo: null as Account | null,
+    balance: BI.from(0),
+    txHash: "",
+  });
+  const [transferTargets, transferTargetsActions] = useList([createTransferTarget()]);
```
PainterPuppets: Introduces unnecessary complexity.

if you want to show fee, you can do so
``` ts
const transfer = () => {
    const unsignedTxSkeleton = ......
   
    setFee(calculateTransactionFee(unsignedTxSkeleton))
    
    // ... other logic
}
```

examples/secp256k1-transfer/src/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+  });
+  const [transferTargets, transferTargetsActions] = useList<TransferTarget>([createTransferTarget()]);
+
+  // this method will be called when you click "Transfer" button
+  const doTransfer = () => {
+    if (!state.accountInfo) {
+      return;
+    }
+
+    transfer({ targets: transferTargets, address: state.accountInfo.address }, state.privateKey).then(
```
PainterPuppets, line 33: The `transfer` function return `fee` field has a bit strange
I understand that if you want the user to confirm the fee before signing then you should separate the creation and signing steps

IronLu233: But it is a example. Abstract is good for development, but not the best for here.
Tells user the transaction fee is from `sum(capacity of inputs) - sum(capacity of outputs)` is more important

PR #395, opened 2022-08-31
feat(ckb-indexer): support indexer rpc v0.4.1

PainterPuppets:
related task: https://github.com/nervosnetwork/lumos/issues/390

### feature:
- [x] support indexer rpc `get_cells`'s `with_data` options 
- [x] support indexer rpc `get_transaction`'s `group_by_transaction` options
- [x] support indexer rpc filter `script_len_range` options

### chore:
- [x] update `ckb-indexer` download script

### test case:
- [x] call `get_cells` with `with_data` option
- [x] call `get_transaction` with `group_by_transaction` option
- [x] call `get_cells` with `script_len_range` filter options
- [x] check `CellCollector` validate & convert scriptLenRange query

### DIFF
--- a/packages/ckb-indexer/src/collector.ts
+++ b/packages/ckb-indexer/src/collector.ts
@@ -104,6 +104,10 @@ export class CKBCellCollector implements BaseCellCollector {
         queries.outputDataLenRange[1]
       );
     }
+    if (queries.scriptLenRange) {
+      utils.assertHexadecimal("scriptLenRange[0]", queries.scriptLenRange[0]);
+      utils.assertHexadecimal("scriptLenRange[1]", queries.scriptLenRange[1]);
+    }
 
     if (queries.outputDataLenRange && queries.data && queries.data !== "any") {
       const dataLen = getHexStringBytes(queries.data);
@@ -153,6 +157,10 @@ export class CKBCellCollector implements BaseCellCollector {
           "0x" + (dataLenRange + 1).toString(16),
         ];
       }
+    }
+
+    if (!this.queries.scriptLenRange && this.queries.type === "empty") {
+      this.queries.scriptLenRange = ["0x0", "0x1"];
     }
   }
 
--- a/packages/ckb-indexer/tests/test_cases.js
+++ b/packages/ckb-indexer/tests/test_cases.js
@@ -8153,6 +8153,14 @@ const cellCollectorTestCases = [
     expectedResult: [],
   },
   {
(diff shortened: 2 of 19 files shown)


packages/ckb-indexer/tests/test_cases.js
```diff
@@ -12,20 +12,22 @@ const lock = {
 const lockScriptWrapper = {
   script: {
     args: "0x7f2b32878434041acd17e396936de4f6ba377cce",
-    codeHash: "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
+    codeHash:
```
zhangyouxin: Looks like that prettier is doing something unexpected

packages/ckb-indexer/src/collector.ts
```diff
@@ -154,6 +158,10 @@ export class CKBCellCollector implements BaseCellCollector {
         ];
       }
     }
+
+    if (!this.queries.scriptLenRange && this.queries.type === "empty") {
```
zhangyouxin, line 162: Not sure if this one is necessary, is there any error if we don't do this change?

zhangyouxin:
Maybe we should add some test cases in `collector.e2e.test.ts` and `transaction_collector.e2e.test.ts` to test `collector.ts` and `transaction_collector.ts`.

packages/ckb-indexer/src/collector.ts
homura: > Not sure if this one is necessary, is there any error if we don't do this change?

this makes filter works in ckb-indexer

packages/ckb-indexer/tests/test_cases.js
homura: @PainterPuppets what has changed in this file? can you please ignore this file from .prettierrc?

PainterPuppets:
> Maybe we should add some test cases in `collector.e2e.test.ts` and `transaction_collector.e2e.test.ts` to test `collector.ts` and `transaction_collector.ts`.

Actually `transaction_collector` and `collector` don't add `with_data` and `group_by_transactions` options, I'm unclear what aspect of the test should be added?

zhangyouxin:
> Actually `transaction_collector` and `collector` don't add `with_data` and `group_by_transactions` options, I'm unclear what aspect of the test should be added?

OK I understand

homura (approved):

PR #400, opened 2022-09-02
fix(common-scripts): avoid run vm1 code in vm0 when deploy

homura:
ckb-vm has upgraded and the vm1 is compatible with vm0, but vm0 is not full supported vm1, we should avoid running vm1's code in vm0


PainterPuppets (approved):

zhangyouxin (approved):
LGTM

PR #403, opened 2022-09-05
feat(codec): molecule friendly error message

### LINKED ISSUE #396: [feature request] validator for codec when packing the object
## Motivation

When I call the `codec.pack`, especially when packing a large or nested object, i'm hard to troubleshoot which part of the object is incorrect, and it would be much easier to debug if the pack could tell me exactly which part is incorrect

## How

```ts
const RGB = array(Uint8, 3)

//       overflow
//         👇
RGB.pack([0xfff, 0xff, 0xff]) // [0] is incorrect, expected ..., actual ...
```

```ts
const RGB = array(Uint8, 3)
const Styled = struct({
  background: RGB,
  borderColor: RGB
}, ['background', 'borderColor']);

Styled.pack({
  //          overflow
  //             👇
  background: [0xfff, 0xff, 0x00],
  borderColor: [0x12, 0x34, 0x56],
}) // error: background.[0] is incorrect, expected ..., actual ...

```

### `.validate` for Codec

Can we add a `validatePack` method to the codec to replace the validators in toolkit

```ts
WitnessArgs.validatePack({...}) // throw an error if the param is invalid
```

https://github.com/nervosnetwork/lumos/blob/d5e7636736e22c3247d36b4160213e74ed23dd0e/packages/toolkit/src/validators.js#L378-L398

### Unpack

For unpack, is there a better way to tell me what went wrong when unpack goes wrong?

IronLu233:
https://github.com/nervosnetwork/lumos/issues/396


packages/codec/tests/molecule.test.ts
```diff
@@ -214,3 +215,225 @@ test("test fixed hex bytes", (t) => {
...
 });
+test("test simple array codec error", (t) => {
+  const codec = array(Uint8, 3);
+  try {
+    codec.pack([0x1, 0xfff, 3]);
+    t.fail();
+  } catch (e) {
+    t.truthy(
+      (e as CodecExecuteError).message.includes(
+        "Expect type Uint8BE at input[1] but got error: Value must be between 0 and 255, but got"
```
zhangyouxin: Uint8 only contains 1 byte, so BE and LE is the same for it. Can you add some other test cases for Uints, like `Uint32LE`

packages/codec/src/base.ts
```diff
@@ -114,3 +114,47 @@ export function createFixedBytesCodec<Unpacked, Packable = Unpacked>(
     }),
   };
 }
+
+export const CODEC_OPTIONAL_PATH = Symbol("?");
```
zhangyouxin: better create an `error.ts`

PainterPuppets:
Can you add description and purpose about this pull request?

packages/codec/tests/molecule.test.ts
```diff
@@ -214,3 +215,225 @@ test("test fixed hex bytes", (t) => {
...
+    t.fail();
+  } catch (e) {
+    t.truthy(
+      (e as CodecExecuteError).message.includes(
+        "Expect type Uint8BE at input.f2 but got error: Value must be between 0 and 255, but got"
+      )
+    );
+  }
+
+  try {
```
PainterPuppets: maybe use `const error = t.throws(...)`, you can search for `t.throws` to find many usage example

IronLu233: https://github.com/avajs/ava-docs/blob/main/zh_CN/readme.md#throwsfunctionpromise-error-message

A good suggestion. we can test the error message in the second parameter.

PainterPuppets (commented):
I'm wondering if it's possible to use the higher-order function to do non-intrusive modifications instead of directly changing the `codec`?

maybe example like that
```
withReadableError(array)(xxxx)
```

packages/codec/tests/molecule.test.ts
IronLu233: OK. I also add some cases for bigger units.

IronLu233:
> I'm wondering if it's possible to use the higher-order function to do non-intrusive modifications instead of directly changing the `codec`?
> 
> maybe example like that
> 
> ```
> withReadableError(array)(xxxx)
> ```

**It is difficult.**

We want track the property access path in codec pack input, but these key only appears in `Array.prototype.map/reduce`
methods.

When code run out of these high order function, we can not get the **current** key anymore. so I think we have to modify some code in codec pack logic.

homura:
> When code run out of these high order function, we can not get the current key anymore. so I think we have to modify some code in codec pack logic.

vote for this. but i want to know it is possible to refactor `createObjectCodec` and `createArrayCodec` to provide the friendly message? because it's not just molecule that can be affected by nested error. for struct/table/array/vector will call the `createObject|ArrayCodec` to report the friendly message

IronLu233:
> > When code run out of these high order function, we can not get the current key anymore. so I think we have to modify some code in codec pack logic.
> 
> vote for this. but i want to know it is possible to refactor `createObjectCodec` and `createArrayCodec` to provide the friendly message? because it's not just molecule that can be affected by nested error. for struct/table/array/vector will call the `createObject|ArrayCodec` to report the friendly message

Maybe ES6 proxy class can implement this feature.

homura:
> Maybe ES6 proxy class can implement this feature.

it can be simpler

```ts
function struct(shape, fields) {
  const codec = createObjectCodec(...)

  return {
    pack(packable) {
      const packedFields = fields.map(field => codec[field].pack(packable[item]))
      return concat(...packedFields) 
    }
    unpack() {
      // ...
    }
  }
}
```

IronLu233:
> > Maybe ES6 proxy class can implement this feature.
> 
> it can be simpler
> 
> ```ts
> function struct(shape, fields) {
>   const codec = createObjectCodec(...)
> 
>   return {
>     pack(packable) {
>       const packedFields = fields.map(field => codec[field].pack(packable[item]))
>       return concat(...packedFields) 
>     }
>     unpack() {
>       // ...
>     }
>   }
> }
> ```

I found a way to track layout struct(`array`, `table`) error without modify it's codec implementations.

This can be in another PR in https://github.com/nervosnetwork/lumos/pull/408

packages/codec/src/error.ts
```diff
@@ -0,0 +1,54 @@
+export const CODEC_OPTIONAL_PATH = "______?______";
```
homura:
```suggestion
export const CODEC_OPTIONAL_PATH = "__molecule_option__";
```

homura:
```suggestion
// lc for lumos codec
export const CODEC_OPTIONAL_PATH = "__lc_option__";
```
i think the `____?____` is too magic when debugging

packages/codec/src/error.ts
```diff
@@ -0,0 +1,54 @@
...
+  constructor(message: string, public expectedType: string) {
+    super(message);
+  }
+}
+
+const CODEC_EXECUTE_ERROR_NAME = "CodecExecuteError";
+export function isCodecExecuteError(
+  error: unknown
+): error is CodecExecuteError {
+  return (error as Error)?.name === CODEC_EXECUTE_ERROR_NAME;
```
homura:
```suggestion
  if (!(error instanceof Error)) return false;
  return error.name === CODEC_EXECUTE_ERROR_NAME;
```

we should avoid using `as` if possible

IronLu233: I agree. Although at here I use optional chaining for preventing access nullable object.

IronLu233:
Seems everything is OK. Can you please help review again? @homura

homura (approved):

PR #408, opened 2022-09-12
feat(codec): add friendly error alternative solution

### LINKED ISSUE #396: [feature request] validator for codec when packing the object

IronLu233:
# Description

Fixes https://github.com/nervosnetwork/lumos/issues/396

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] Test for simple codec type
- [x] Test for nested codec type
- [ ] Test for createObjectCodec or createArrayCodec

### DIFF
--- a/packages/codec/src/base.ts
+++ b/packages/codec/src/base.ts
@@ -114,3 +114,90 @@ export function createFixedBytesCodec<Unpacked, Packable = Unpacked>(
     }),
   };
 }
+
+const LAST_ACCESS_PATH = Symbol("PACKABLE_ACCESS_PATH");
+
+export class CodecBaseParseError extends Error {
+  constructor(message: string, public expectedType: string) {
+    super(message);
+  }
+}
+
+class CodecPackableProxyHandler<T extends Record<any, unknown> | any[]>
+  implements ProxyHandler<T>
+{
+  constructor(
+    private previousAccessPaths: PropertyKey[] = [],
+    private onAccess?: (paths: PropertyKey[]) => unknown
+  ) {}
+
+  lastAccess: PropertyKey[] = [];
+
+  get(target: T, p: string | symbol): any {
+    if (p === LAST_ACCESS_PATH) {
+      return this.lastAccess;
+    }
+
+    const value = target[p as keyof T];
+
+    // property may be an array's index, but proxy will automatically parse it to string. so we need to convert it back to number
+    const prop =
+      Array.isArray(target) && /^\d+$/.test(p as string) ? Number(p) : p;
+    const paths = [...this.previousAccessPaths, prop];
+    const onAccess =
+      this.onAccess ||
+      ((paths: PropertyKey[]) => {
+        this.lastAccess = paths;
+      });
+    if (!(p in (Reflect.getPrototypeOf(target) || Object.create(null)))) {
+      onAccess(paths);
+    }
+    if (Array.isArray(value) || (value !== null && typeof value === "object")) {
+      return new Proxy(
+        value as any,
+        new CodecPackableProxyHandler(paths, onAccess)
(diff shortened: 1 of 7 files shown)


packages/codec/src/base.ts
```diff
@@ -114,3 +114,90 @@ export function createFixedBytesCodec<Unpacked, Packable = Unpacked>(
     }),
   };
 }
+
+const LAST_ACCESS_PATH = Symbol("PACKABLE_ACCESS_PATH");
```
homura, line 118: as a lib, Lumos should have as minimal dependencies and state as possible

we should avoid using symbol, because users may use multiple versions of the codec, which will make the old codec and the new version not work properly

for example, if a library uses `codec@0.1` to create a blockchain.mol codec and the user uses `codec@0.2`, this will make `codec0_2HighOrderCodec(codec0_1.someCodec)` not work properly because they are not using the same symbol

packages/codec/src/base.ts
```diff
@@ -114,3 +114,90 @@ export function createFixedBytesCodec<Unpacked, Packable = Unpacked>(
...
+        return codec.pack(packableProxy);
+      } catch (e) {
+        const lastAccessPaths: PropertyKey[] = packableProxy[LAST_ACCESS_PATH];
+        const paths = (lastAccessPaths as (number | string)[]).reduce(
+          (acc, cur) => {
+            return acc + (typeof cur === "number" ? `[${cur}]` : `.${cur}`);
+          },
+          "input"
+        );
+        if (e instanceof CodecBaseParseError) {
```
homura, line 193: same with above, we should avoid using `instanceof` to check error type here

packages/codec/src/base.ts
```diff
@@ -114,3 +114,90 @@ export function createFixedBytesCodec<Unpacked, Packable = Unpacked>(
...
+
+class CodecPackableProxyHandler<T extends Record<any, unknown> | any[]>
+  implements ProxyHandler<T>
+{
+  constructor(
+    private previousAccessPaths: PropertyKey[] = [],
+    private onAccess?: (paths: PropertyKey[]) => unknown
+  ) {}
+
+  lastAccess: PropertyKey[] = [];
```
homura, line 134: so the `lastAccess` is a stateful prop here? if a developer create a custom codec, it is working correctly? like this

```ts
const codec = createCodec({
  pack(obj) {
     const a = obj.a;
     const b = obj.b;

     return concat(bytify(a) /* error here */, bytify(b));
  }
})
```

PR #413, opened 2022-09-15
feat: molecule schema parser

### LINKED ISSUE #402: [feature request] Parsing molecule schema to codec directly
## Motivation

As a dApp developer, i often explore some transactions in the ckb explorer to learn how a dApp works, but the `script.args`/`cell.data`/`transaction.witness` in the ckb explorer is serialized and represented as a hex string, most of this data is serialized from molecule. It would be very helpful to have a tool that would allow me to input the schema and these hex strings for decode, so I could understand the transaction

## How

We can offer a new API

```ts
type CodecRecord = Record<string, AnyCodec>;

declare function parseMolecule<
  Ref extends CodecRecord = CodecRecord, 
  Ret extends CodecRecord = CodecRecord
>(
  // molecule schema string
  mol: string, 
  // referenceable codec for mol, e.g. codecs.numbers, base.blockchain
  options?: { refs: Ref }
): Ret:
```

```ts
// language=molecule
const mol = `
array RGB (Uint8, 3);

struct Styled {
  background: RGB,
  color: RGB,
}
`;

import { number } from '@ckb-lumos/codec'

const codecs = parseMolecule(mol, { references: ...number /* referenceable codecs */})
```

After implement, we can offer a toolkit on the Lumos website to helps user to decode molecule binary

image

zhangyouxin:
# Description

This PR provides an easy way to translate molecule schema to codecs. More deep motivations please refer to #402 

## Usage

```ts
// molecule schema
const mol = `array RGB [Uint8; 3];`

// translate schema to codecs
import { createParse } from "@ckb-lumos/molecule"
const tokens = createParser().parse(mol, { refs: { Uint8: number.Uint8 }})

// decode
codecMap["RGB"].unpack("0x010203")
```

Resolved #402 

This PR introduces a new npm package: `@ckb-lumos/molecule` 

demo website: https://lumos-website-git-fork-zhangyouxin-molecule-cryptape.vercel.app/tools/molecule-parser
demo usage: https://github.com/zhangyouxin/lumos/blob/b3c9ad0486a7b945a82b6ff67c7b15a471d945ea/website/src/components/molecule-parser/Molecule.tsx#L55-L65

## Type of change

- New feature (non-breaking change which adds functionality)
- Documentation or Website

# How Has This Been Tested?

- [x] Unit Test for parsing schema to token
- [x] Unit Test for parsing token to codecs


homura:
found a preview is enabled at https://mol-parser-ts.vercel.app/

packages/codec/package.json
```diff
@@ -30,7 +30,8 @@
     "release": "npm publish"
   },
   "dependencies": {
-    "@ckb-lumos/bi": "^0.19.0-alpha.1"
+    "@ckb-lumos/bi": "^0.19.0-alpha.1",
```
homura:
```suggestion
    "@ckb-lumos/bi": "^0.19.0-alpha.3",
```

.eslintrc.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
   ],
   "rules": {
     "no-var": "error",
+    "no-case-declarations" : "warn",
```
homura: why we need to modify the lint rule? here is a tip for how to refactor https://eslint.org/docs/latest/rules/no-case-declarations

zhangyouxin: got

packages/codec/package.json
zhangyouxin: It will be fixed when merge develop into this PR

homura:
hi @zhangyouxin, can you please split the PR into two, one for the molecule parser and the other for the integration of the parser with the parser

packages/codec/src/utils.ts
```diff
@@ -3,6 +3,80 @@ import {
...
+export const deepHexifyBI = deepTranslateBI("toHexString");
+
+/**
+ * Unpack result is either number, string, object, or BI
+ * convert { field: BI } to { field: string } in order to compare unpack results in tests
+ *
+ * e.g. { capacity: BI.from(10) } ==> { capacity: "10" }
+ * @param data
+ */
+export const deepDecimalizeBI = deepTranslateBI("toString");
```
homura: looks a bit inflexible, can we implement a transformer like this

```ts


type PredMapTo<T, From, To> = T extends From ? To : T;
declare function predMap<T, From, To>(
  origin: T, 
  pred: (x: unknown) => x is from, 
  map: (from: From) => To
): PredMapTo<T, From, To>;

type ObjectLike = Record<string, any>;
type DeepPredMapObjectTo<
  T extends ObjectLike,
  From /* map from */,
  To /* map to */
> = {
  [K in keyof T]: T[K] extends ObjectLike
    ? DeepPredMapObjectTo<T[K], From, To>
    : PredMapTo<T[K], From, To>;
};

declare function function deepMapObject<T extends ObjectLike, From, To>(
  obj: T,
  pred: (x: unknown) => x is From,
  mapping: (from: From) => To
): DeepPredMapObjectTo<T, From, To>;

// array is same with above
```

homura:
maybe we can make the API more simple to reduce the template code

```ts
const codecs = parseMolecule(`
array Uint8 [byte; 1];
array RGB [Uint8; 3];
`, {
  refs: {
    Uint8: codec.number.Uint8,
  }
})

codecs.RGB.unpack("0xaabbcc")
```

Or 

```ts

const codecs = parseMolecule(`
array Uint8 [byte; 1];
array RGB [Uint8; 3];
`, {
  refs: {
    `array Uint8 [byte; 1];`: codec.number.Uint8,
  }
})

codecs.RGB.unpack("0xaabbcc")
```

all the refs codec can be custom like this 

```ts

declare function enhanceUnpack<
  C extends AnyCodec, 
  E /* enhanced unpacked */
>(codec: C, enhancer: (x: UnpackResult<C>) => E): Codec<
  PackResult<C>, 
  E, 
  PackParam<C>, 
  UnpackParam<C>
>

const humanizeBigInteger = (x: BIish) => numeral.format(BI.from(x).toString(), '0,0');
const refs =  { 
  Uint8: enhanceUnpack(Uint8,  humanizeBigInteger), 
  ..., 
  Uint512: enhanceUnpack(Uint512, humanizeBigInteger) 
}

// so that we can

parseMolecule(moleculeDefs, { refs }); // capacity: 10,000,000,000

```

website/src/constants/builtinCodecs.ts
```diff
@@ -1,47 +1,65 @@
-import { blockchain } from '@ckb-lumos/base';
+import { AnyCodec, number } from "@ckb-lumos/codec"
+import { blockchain } from "@ckb-lumos/base"
 import { CodecMap } from "@ckb-lumos/molecule"
-import { AnyCodec } from '@ckb-lumos/codec';
+import { BIish, BI } from "@ckb-lumos/bi/lib"
+import numeral from "numeral"
+
+function enhanceUnPackBIish(codec: AnyCodec, afterUnpack: (arg: BIish) => string) {
```
homura: unpack is a word, so that we can rename `enhanceUnPackBIish` to `enhanceUnpackBIish`

website/src/components/molecule-parser/DataInut.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const [result, setResult] = useState<BITranslatedUnpackType>(undefined)
+  const [errorMsg, setErrorMsg] = useState<string>("")
+  const handleDecode = () => {
+    if (!props.codec) {
+      setErrorMsg("please select codec")
+      return
+    }
+    try {
+      const result = props.codec.unpack(formatInput(inputData))
+      setResult(deepStringifyNumber(result))
```
homura: why do we still need a `deepStringifyNumber` here? the `props.codec.unpack` is not a plain JSON here?

packages/molecule/src/utils.ts
```diff
@@ -0,0 +1,16 @@
+import { MolType, MolTypeMap } from "./type";
+
+// TODO: assert not null/undefined
+export function nonNull(data: unknown): void {
+  if (!data) {
+    throw new Error(`${data} does not exist.`);
+  }
+}
```
homura:
```suggestion
export function nonNull<T>(data: T): asserts data is NonNullable<T> {
  if (data == null) throw new Error('NonNullable') 
}
```

website/src/constants/builtinCodecs.ts
```diff
@@ -0,0 +1,44 @@
...
+import numeral from "numeral"
+
+function enhanceUnpackBIish(codec: AnyCodec, afterUnpack: (arg: BIish) => string) {
+  return {
+    ...codec,
+    unpack: (packed: Uint8Array) => afterUnpack(codec.unpack(packed)),
+  }
+}
+
+const humanizeBigInteger = (x: BIish) => numeral(BI.from(x).toString()).format("0,0")
```
homura: the comment is just an example to show how to humanize, as a JSON, i think we don't need to format it

packages/base/tests/deserialize.test.ts
```diff
@@ -0,0 +1,319 @@
...
+import { HashType } from "@ckb-lumos/base";
+import { DepType } from "./../src/api.d";
+import test from "ava";
+import * as blockchain from "../src/blockchain";
+import type * as api from "../src/api";
+import { BI } from "@ckb-lumos/bi";
+
+test.before(() => {
+  JSBI.prototype["valueOf"] = () => 0;
+});
```
homura: any reason for the overriding here?

zhangyouxin: The behaviour of `deepEqual` in ava test environment is to recursively call `valueOf` of every field in the values to be compared. 

Since `BI` in `@ckb-lumos/bi` has a `JSBI` instance, `jsbi.valueOf` will be called when calling `t.deepEqual(BI.from(1), BI.from(1))`, and that will always throw an Error: https://github.com/GoogleChromeLabs/jsbi/blob/main/lib/jsbi.ts#L80

homura: i see, but the `JSBI.prototype["valueOf"] = () => 0;` may cause all `t.deepEquals` to be true when comparing `JSBI`/`BI`. It would be better to use `JSBI.prototype.valueOf = function() { return this.toString(); }`

zhangyouxin: yes i have thought about this issue too, but according to test result indicates it won't, because `valueOf` called in BI has already done the comparison. so I figure it's not necessary to do it the second time.

zhangyouxin: but anyway you are right, changing to `JSBI.prototype.valueOf = function() { return this.toString(); }` makes it more readable and comprehensive.

homura: Also, I would say that this code is confusing, and we should comment the reason for doing this

website/src/components/molecule-parser/DataInut.tsx
```diff
@@ -0,0 +1,88 @@
...
+}
+
+const isBlank = (data: UnpackType): boolean => {
+  if (!data) {
+    return true
+  }
+  return false
+}
+
+export const DataInput: React.FC<Props> = (props) => {
```
PainterPuppets: Is the file name typo, it should be `DataInput.tsx`

website/src/components/molecule-parser/DataInut.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const handleDecode = () => {
+    if (!props.codec) {
+      setErrorMsg("please select codec")
+      return
+    }
+    try {
+      const result = props.codec.unpack(formatInput(inputData))
+      setResult(result)
+      setErrorMsg("")
+    } catch (error: any) {
```
PainterPuppets: maybe can using `catch (error: Error)` directly?

homura:
the low coverage causes the CI to fail, can you check the `nearley.ts`

packages/lumos/src/index.ts
```diff
@@ -46,3 +46,4 @@ export { CellCollector, Indexer } from "@ckb-lumos/ckb-indexer";
 export * as helpers from "@ckb-lumos/helpers";
 export * as commons from "@ckb-lumos/common-scripts";
 export { BI } from "@ckb-lumos/bi";
+export * as molecule from "@ckb-lumos/molecule";
```
homura:
```suggestion

```

the `molecule` may be confused with `codec.molecule`, we should find another better name

homura (approved):
LGTM

PR #427, opened 2022-09-25
feat(helpers): add `validateP2PKHMessage` function

### LINKED ISSUE #397: [idea] Transaction validator for P2PKH signing message
## Motivation

The CKB contract is very flexible and can theoretically support any form of HMAC, so it can support signature verification on any blockchain in CKB, therefore, CKB can use other blockchain wallets as CKB's wallet, e.g. [Ethereum](https://github.com/nervosnetwork/lumos/tree/develop/examples/omni-lock-metamask) / [Cardano](https://github.com/nervosnetwork/lumos/tree/develop/examples/cardano-lock-namiwallet) etc. 

The user experience of these wallets is excellent, the transaction will be displayed when signing. For example, when we transfer ERC20, MetaMask will show the recipient address, type of the token and token amount, this is because MetaMask is able to recognize ERC20 transactions. However, MetaMask doesn't recognize CKB's transactions, so we can only use an API like `personal_sign` to sign the hashed transaction. The hashed transactions are unreadable for humans.

## Idea

We can provide a `validateSigningMessage` method to check `diget == hash(tx | witness)`, this helps the user to verify that a transaction is tampered.

```ts

type HashAlgorithm = 
  | 'ckb-blake2b-256'
  | 'ckb-blake2b-160' 
  | 'sha256' 
  | 'ripemd-160' 
  | 'sha3'
  | '...'
  | ((message: BytesLike) => Uint8Array);

interface Preimage {
  rawTransaction: RawTransaction;
  // witness
  additionalPayload: BytesLike;
}

declare function validateSigningMessage(
  preimage: Preimage,
  messageForSigning: BytesLike,
  options?: {
    hashAlgorithm: HashAlgorithm;
  }
): boolean

```

### Wallet gateway

```ts
// TODO a wallet gateway
type SignatureAlgorithms = 
  | 'eth_personal_sign' 
  | 'eth_signTypedData_v4' 
  | 'cardano_personal_sign'
  | '...';

type PubkeyLike =
  // public key 
  | BytesLike 
  // public key hash
  | { hashAlgorithm: HashAlgorithm, pubkeyHash: BytesLike }
```

### Todo - Explain `additionalPayload`

IronLu233:
# Description

Fixes #397
provide a `validateP2PKHMessage ` method to check diget == hash(tx | witness), this helps the user to verify that a transaction is tampered.
### Usage
``` typescript
const rawTx = /* some create transaction operation */;
const tx = prepareSigningEntries(rawTx);
const txMessageDigest = prepareSigningEntries.signingEntries.get(0).message;
// ------- this process may modify transaction message digest
// return the validate result
validateP2PKHMessage(txMessageDigest, tx,  hashContentExceptRawTx);
```
## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] Test simple validation
- [x] Test multiple lock args
- [x] Test extra custom witnesses

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/packages/base/src/utils.d.ts
+++ b/packages/base/src/utils.d.ts
@@ -4,6 +4,7 @@ import { Hash, HexNumber, HexString } from "./primitive";
 import { BytesLike } from "@ckb-lumos/codec";
 
 export class CKBHasher {
+  constructor(options?: { outLength?: number });
   update(data: string | ArrayBuffer): this;
 
   digestHex(): Hash;
@@ -12,26 +13,26 @@ export class CKBHasher {
 export function ckbHash(data: BytesLike): Hash;
 
 /**
- * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide} 
+ * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide}
  * convert bigint to BigUInt64 little-endian hex string
  * @param num
  */
 export function toBigUInt64LE(num: BIish): HexString;
 
 /**
- * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide} 
+ * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide}
  * convert BigUInt64 little-endian hex string to bigint
  *
  * @param hex BigUInt64 little-endian hex string
  */
 export function readBigUInt64LE(hex: HexString): bigint;
 /**
- * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide} 
+ * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide}
  */
 export function readBigUInt64LECompatible(hex: HexString): BI;
 
 /**
- * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide} 
+ * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide}
  * convert bigint to BigUInt128 little-endian hex string
  *
  * @param u128
@@ -39,14 +40,14 @@ export function readBigUInt64LECompatible(hex: HexString): BI;
 export function toBigUInt128LE(u128: BIish): string;
 
 /**
- * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide} 
+ * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide}
  * convert BigUInt64 little-endian hex string to bigint
  *
  * @param leHex BigUInt128 little-endian hex string
  */
 export function readBigUInt128LE(leHex: HexString): bigint;
 /**
- * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide} 
+ * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide}
  */
 export function readBigUInt128LECompatible(leHex: HexString): BI;
 
--- a/packages/base/src/utils.d.ts
+++ b/packages/base/src/utils.d.ts
@@ -3,8 +3,12 @@ import { Script, Input } from "./api";
 import { Hash, HexNumber, HexString } from "./primitive";
 import { BytesLike } from "@ckb-lumos/codec";
 
+type CKBHasherOptions = {
+  outLength?: number;
+};
+
 export class CKBHasher {
-  constructor(options?: { outLength?: number });
+  constructor(options?: CKBHasherOptions);
   update(data: string | ArrayBuffer): this;
 
   digestHex(): Hash;
--- a/packages/base/src/utils.js
+++ b/packages/base/src/utils.js
@@ -7,8 +7,15 @@ const blockchain = require("./blockchain");
 
 const { bytify, hexify, bytifyRawString } = bytes;
 class CKBHasher {
-  constructor() {
-    this.hasher = blake2b(32, null, null, bytifyRawString("ckb-default-hash"));
+  constructor(options = {}) {
+    const { outLength = 32 } = options;
+    this.outLength = outLength;
+    this.hasher = blake2b(
+      outLength,
+      null,
+      null,
+      bytifyRawString("ckb-default-hash")
+    );
   }
 
   update(data) {
@@ -17,7 +24,7 @@ class CKBHasher {
   }
 
   digestHex() {
-    const out = new Uint8Array(32);
+    const out = new Uint8Array(this.outLength);
     this.hasher.digest(out);
     return hexify(out.buffer);
   }
--- a/packages/helpers/package.json
+++ b/packages/helpers/package.json
@@ -42,10 +42,8 @@
     "@ckb-lumos/config-manager": "^0.19.0-alpha.3",
     "@ckb-lumos/toolkit": "^0.19.0-alpha.3",
     "@types/blake2b": "^2.1.0",
-    "@types/lodash.pick": "^4.4.7",
     "blake2b": "^2.1.3",
     "bech32": "^2.0.0",
-    "lodash.pick": "^4.4.0",
     "immutable": "^4.0.0-rc.12"
   },
   "devDependencies": {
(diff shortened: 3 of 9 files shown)


packages/helpers/src/validate.ts
```diff
@@ -0,0 +1,116 @@
...
+ * @param extraData the extra data for validate
+ * @param hashAlgorithm hash algorithm for signing. @default "ckb-blake2b-256"
+ * @returns the validate result. unless all messages equals, it will return false.
+ */
+export function validateTransaction(
+  messagesForSigning: BytesLike[],
+  rawTransaction: TransactionSkeletonType,
+  extraData: BytesLike[],
+  hashAlgorithm: HashAlgorithm = "ckb-blake2b-256"
+): boolean {
```
homura: 1. propose to change `validateTransaction` to `validateP2PKHMessage` since `validateTransaction` is unambiguous, it will make users think it is validating a transaction is correct
2. propose to change `extraData` to `hashContentExceptRawTransaction` since `extraData` here may confuse with [extra_hash](https://github.com/nervosnetwork/ckb/blob/develop/util/types/schemas/blockchain.mol#L80), maybe we can use `witnessPartForHash`

packages/helpers/src/validate.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+}
+
+function getWitnessLength(witnesses: Uint8Array) {
+  return number.Uint64.pack(witnesses.length).buffer;
+}
+
+/**
+ * Validate an transaction.
+ * @param messagesForSigning the message for signing. Each item in it means hash(rawTransaction | extraData).
```
homura: the digest message for signing.

packages/helpers/src/validate.ts
```diff
@@ -0,0 +1,116 @@
...
+ * @param rawTransaction raw transaction object
+ * @param extraData the extra data for validate
```
homura: content to be hashed other than `rawTransaction`, is generally processed witness

packages/helpers/src/validate.ts
IronLu233: `hashContentExceptRawTx` is shorter than `hashContentExceptRawTransaction`.

homura:
This PR is a fix for #397, can you please describe clearly the purpose of this PR?

packages/helpers/src/validate.ts
```diff
@@ -0,0 +1,116 @@
...
+    .update(
+      RawTransaction.pack(
+        pick(tx, [
+          "cellDeps",
+          "headerDeps",
+          "inputs",
+          "outputs",
+          "outputsData",
+          "version",
+        ])
```
homura: why we need `pick` here?

IronLu233: If we don't pick the fields, Typescript type checking will refuse the parameter.

packages/helpers/package.json
```diff
@@ -41,6 +41,8 @@
     "@ckb-lumos/bi": "^0.19.0-alpha.3",
     "@ckb-lumos/config-manager": "^0.19.0-alpha.3",
     "@ckb-lumos/toolkit": "^0.19.0-alpha.3",
+    "@types/blake2b": "^2.1.0",
+    "blake2b": "^2.1.3",
```
homura: we can use `CkbHasher` or `ckbHash` in `@ckb-lumos/base`

packages/helpers/src/validate.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @param messagesForSigning the message digest for signing. means hash(rawTransaction | extraData).
+ * @param rawTransaction raw transaction object
+ * @param hashContentExceptRawTx content to be hashed other than rawTransaction, is generally processed witness
+ * @param hashAlgorithm hash algorithm for signing. Default is `"ckb-blake2b-256"`
+ * @returns the validate result. unless all messages equals, it will return false.
+ */
+export function validateP2PKHMessage(
+  messagesForSigning: BytesLike,
+  rawTransaction: TransactionSkeletonType,
+  hashContentExceptRawTx: BytesLike,
```
zhangyouxin, line 38: Is it possible to merge `hashContentExceptRawTx` and `rawTransaction` into one parameter, cause everything it needs to calculate message is in transactionSkeleton.

BTW, this implementation seems unable to handle a case that the tx has multiple witness like: `[witness1, witness2]`

IronLu233: > BTW, this implementation seems unable to handle a case that the tx has multiple witness like: [witness1, witness2]

In current version. support it is not in current stage. because the hash algorithm may be multiple in this stage(We want this function not only support CKB, but also support other transaction

IronLu233: > BTW, this implementation seems unable to handle a case that the tx has multiple witness like: [witness1, witness2]
One signing entries just have one message and `hashContentExceptRawTx`

packages/helpers/package.json
IronLu233: I think we CAN'T
because in https://github.com/ckb-js/lumos/blob/e33aaa10d831edd58e904cb2215ea3148c37fba3/packages/base/src/utils.js#L11
the CKBHasher defined as
```
 this.hasher = blake2b(32, null, null, bytifyRawString("ckb-default-hash"));
```
we need change blake2b in differient hash algorithms.

packages/base/src/utils.d.ts
```diff
@@ -4,6 +4,10 @@ import { Hash, HexNumber, HexString } from "./primitive";
 import { BytesLike } from "@ckb-lumos/codec";
 
 export class CKBHasher {
+  /**
+   * @param outLength the length of the output hash in bytes. default is `32`
+   */
+  constructor(outLength?: number);
```
PainterPuppets:
```ts
interface CKBHasherOptions {
    outLength?: number;
}
constructor(options?: CKBHasherOptions);
```

packages/base/src/utils.js
```diff
@@ -7,8 +7,14 @@ const blockchain = require("./blockchain");
 
 const { bytify, hexify, bytifyRawString } = bytes;
 class CKBHasher {
-  constructor() {
-    this.hasher = blake2b(32, null, null, bytifyRawString("ckb-default-hash"));
+  constructor(outLength = 32) {
```
PainterPuppets:
```ts
constructor(options = {}) {
    const { outLength = 32 } = options;
    ...
}
```

PainterPuppets (changes requested):
If the first parameter is optional, then if a parameter is added later it will become breakchange

packages/helpers/src/validate.ts
```diff
@@ -0,0 +1,63 @@
...
+  const witnessLength = getWitnessLength(witness);
+  hasher.update(txHash);
+  hasher.update(witnessLength);
+  hasher.update(witness);
+  const actual = hasher.digestHex();
+  if (actual !== message) {
+    return false;
+  }
+
+  return true;
```
homura:
```suggestion
  const hasher = createHasher(hashAlgorithm);
  hasher.update(hashContentExceptRawTx);
  
  return bytes.equal(messagesForSigning, hasher.digestHex())
```


here is a [demo](https://github.com/homura/hello-lumos/blob/wallet-gateway-poc/tutorial/wallet-gateway/gateway.ts) for `hashContentExceptRawTx`, the witness can be hashed more than one way, so the `hashContentExceptRawTx` may need to generate by `prepareSigningEntries`.

packages/helpers/src/validate.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Validate a P2PKH(Pay to public key hash) message
+ * @param messagesForSigning the message digest for signing. means hash(rawTransaction | extraData).
+ * @param rawTransaction raw transaction object
+ * @param hashContentExceptRawTx content to be hashed other than rawTransaction, is generally processed witness
+ * @param hashAlgorithm hash algorithm for signing. Default is `"ckb-blake2b-256"`
+ * @returns the validate result. unless all messages equals, it will return false.
+ */
+export function validateP2PKHMessage(
+  messagesForSigning: BytesLike,
+  rawTransaction: TransactionSkeletonType,
```
PainterPuppets: change to
`txSkeleton: TransactionSkeletonType`
or
`rawTransaction: RawTransaction`

packages/helpers/src/validate.ts
```diff
@@ -0,0 +1,57 @@
...
+  const txHash = rawTxHasher.update(RawTransaction.pack(tx)).digestHex();
+
+  const hasher = createHasher(hashAlgorithm);
+  const message = hexify(messagesForSigning);
+  const witness = bytify(hashContentExceptRawTx);
+  hasher.update(txHash);
+  hasher.update(witness);
+  const actual = hasher.digestHex();
+  if (actual !== message) {
+    return false;
```
IronLu233: Foolish `return false`

homura (approved):
LGTM

PR #432, opened 2022-09-27
feat: support timeout in rpc

zhangyouxin:
# Description

Currently there is no timeout for the lumos rpc methods:

```ts
import { RPC } from '@ckb-lumos/rpc'
const rpc = new RPC('YOUR RPC URL')
// if there is something wrong with network or rpc server causing it stuck
// the rpc requests will never return 
await rpc.getBlock("0x1")
```

This pr fixes this by adding a defaut timeout(30 seconds) to the rpc methods and provides a parameter on RPC constructor which user can customize timeout configuration:

```ts
import { RPC } from '@ckb-lumos/rpc'
const rpc = new RPC('YOUR RPC URL', { timeout: 1 }) // milliseconds

// throws time exceeded error
await rpc.getBlock("0x1")
```

## Type of change

- Refactor (non-breaking change)

# How Has This Been Tested?

- An example is provided to test this feature.
- Added test cases

### DIFF
--- a/packages/rpc/example/mockRpcServer.ts
+++ b/packages/rpc/example/mockRpcServer.ts
@@ -1,0 +1,25 @@
+import express from 'express'
+
+const app = express();
+
+app.use(
+  express.urlencoded({
+    extended: true,
+  })
+);
+
+app.use(express.json());
+
+app.post('/', (req, res) => {
+  setTimeout(() => {
+    console.log(req.body);
+    res.send({
+      "jsonrpc": "2.0",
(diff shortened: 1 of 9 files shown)


packages/rpc/example/mockRpcServer.ts
```diff
@@ -0,0 +1,25 @@
+import express from 'express'
```
homura, line 1: we should put the case in to test case, https://github.com/nock/nock

zhangyouxin: done

IronLu233 (approved):

homura (approved):
LGTM

PR #442, opened 2022-10-12
feat(bi): add support for unit conversions to BI

### LINKED ISSUE #420: `parseUnits` and `formatUnits` for BI

The execution environment of contracts in blockchain usually requires determinism, therefore, the execution environment of general contracts does not directly provide floating point calculation, and for the user side of dApp, floating point is a common scenario, generally we will describe the balance of CKB will use CKB as the unit instead of Shannon, generally we will use 1 CKB instead of saying 100,000,000 shannon (1 ckb = 1e8 shannon)

We need to convert between CKB and shannon in a dApp, if Lumos [BI)(https://github.com/ckb-js/lumos/tree/develop/packages/bi) can provide support for floating point numbers, then it will become more convenient for BI to use

we can expose some API like ethers [utility](https://docs.ethers.io/v5/api/utils/display-logic/) in `@lumos-ckb/bi`

```ts
type Unit = 'shannon' | 'ckb' | number;
declare function parseUnit(str: string, unit: Unit): BI;
declare function formatUnit(bi: BIish, unit: Unit): string;

parseUnit('1', 'ckb') // == BI.from(1e8)
parseUnit('100000000', 'shannon') == BI.from(1e8)
formatUnit(123000000, 'ckb') == '1.23'
```

fghdotio:
# Description

Fixes #420 

## Type of change

- [x] New feature (non-breaking change which adds functionality) that eases the process of unit conversion
  - `function formatUnit(value: BIish, unit: Unit): string`
  -  `function parseUint(value: string, unit: Unit): BI`

### DIFF
--- a/packages/bi/src/index.ts
+++ b/packages/bi/src/index.ts
@@ -185,3 +185,154 @@ export function toJSBI(value: BIish): JSBI {
     return JSBI.BigInt(value.toString());
   }
 }
+
+export type Unit = "shannon" | "ckb" | number;
+
+const validUnitNames = ["shannon", "ckb"];
+
+export const maxDecimals = 8;
+
+const zeros = Array(maxDecimals).fill("0").join("");
+
+const negativeOne = BI.from(-1);
+
+export function formatUnit(value: BIish, unit: Unit): string {
+  const decimals = parseDecimals(unit);
+  return formatFixed(value, decimals);
+}
+
+export function parseUint(value: string, unit: Unit): BI {
+  const decimals = parseDecimals(unit);
+  return parseFixed(value, decimals);
+}
+
+function formatFixed(value: BIish, decimals: number): string {
+  if (!isValidDecimalSize(decimals)) {
+    throw new Error(
+      `decimal size must be within the range of [0-${maxDecimals}]`
+    );
+  }
+
+  const multiplier = "1" + zeros.substring(0, decimals);
+
+  value = BI.from(value);
+  const isNegative = value.isNegative();
+  if (isNegative) {
+    value = value.mul(negativeOne);
+  }
+
+  const wholePart = value.div(multiplier).toString();
+  let result = wholePart;
+  if (multiplier.length > 1) {
+    let decimalPart = value.mod(multiplier).toString();
+    while (decimalPart.length < multiplier.length - 1) {
+      decimalPart = "0" + decimalPart;
+    }
+    // remove trailing zeros
+    decimalPart = decimalPart.match(/^([0-9]*[1-9]|0)(0*)/)![1];
+    result += "." + decimalPart;
+  }
+
+  if (isNegative) {
+    result = "-" + result;
+  }
+
+  return result;
+}
+
+function parseFixed(value: string, decimals: number): BI {
+  if (!isValidDecimalSize(decimals)) {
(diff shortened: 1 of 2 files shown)


packages/bi/src/index.ts
```diff
@@ -185,3 +185,154 @@ export function toJSBI(value: BIish): JSBI {
...
+const zeros = Array(maxDecimals).fill("0").join("");
+
+const negativeOne = BI.from(-1);
+
+export function formatUnit(value: BIish, unit: Unit): string {
+  const decimals = parseDecimals(unit);
+  return formatFixed(value, decimals);
+}
+
+export function parseUint(value: string, unit: Unit): BI {
```
homura:
```suggestion
export function parseUnit(value: string, unit: Unit): BI {
```

packages/bi/src/index.ts
```diff
@@ -185,3 +185,154 @@ export function toJSBI(value: BIish): JSBI {
     return JSBI.BigInt(value.toString());
   }
 }
+
+export type Unit = "shannon" | "ckb" | number;
+
+const validUnitNames = ["shannon", "ckb"];
+
+export const maxDecimals = 8;
```
homura: maybe the `maxDecimals` check is unnecessary, although the decimals of CKB is 8, `BI` is sometimes used to handle sUDT (ERC20-like assets), which may have decimals of 18 or greater, so we should not limit the decimals to 8

packages/bi/src/index.ts
fghdotio: typo fixed

packages/bi/src/index.ts
fghdotio: Good point, agreed. The restriction previously imposed on the maximum size of decimals was lifted.

packages/bi/tests/index.test.ts
```diff
@@ -211,3 +211,90 @@ test("from JSBI", (t) => {
...
+    { value: "1", unit: "ckb", result: BI.from(1e8) },
+    { value: "-01.00", unit: "ckb", result: BI.from(-1e8) },
+    { value: "-0.04", unit: "ckb", result: BI.from(-4e6) },
+    { value: "123.321", unit: "ckb", result: BI.from(123321e5) },
+    { value: "0.00000001", unit: "ckb", result: BI.from(1) },
+    { value: "100000000", unit: "shannon", result: BI.from(1e8) },
+    { value: "0.00000001", unit: 8, result: BI.from(1) },
+    { value: "-0.0000001", unit: 8, result: BI.from(-10) },
+    { value: "0.0000001", unit: 7, result: BI.from(1) },
+    { value: "1.1", unit: 18, result: BI.from(11e17) },
```
homura:
```suggestion
    { value: "1.1", unit: 18, result: BI.from(10).pow(17).mul(11) },
```

11e17 is greater than [MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), although it would not be wrong in this case, it is not really strict enough to do so

fghdotio: insightful input, appreciated it.

homura (approved):
LGTM

PR #447, opened 2022-10-20
refactor(examples): replace omni lock hash code in omnilock metamask example

IronLu233:
# Description

In current omnilock metamask example, it use a custom omnilock cell dependency.

https://github.com/ckb-js/lumos/blob/develop/examples/omni-lock-metamask/lib.ts#L11
``` typescript
export const CONFIG = config.createConfig({
  PREFIX: "ckt",
  SCRIPTS: {
    ...config.predefined.AGGRON4.SCRIPTS,
    // for more about Omni lock, please check https://github.com/XuJiandong/docs-bank/blob/master/omni_lock.md
    OMNI_LOCK: {
      CODE_HASH: "0x79f90bb5e892d80dd213439eeab551120eb417678824f282b4ffb5f21bad2e1e",
      HASH_TYPE: "type",
      TX_HASH: "0x9154df4f7336402114d04495175b37390ce86a4906d2d4001cf02c3e6d97f39c",
      INDEX: "0x0",
      DEP_TYPE: "code",
    },
  },
});
```
But in newer version, we have a predefined omnilock in `config.predefined.SCRIPTS.OMNILOCK`. and we have a new method named `createOmnilockScript` to get omnilock script more convenient.

This PR update the implementation of the example

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] Manual test in codesandbox

### DIFF
--- a/examples/omni-lock-metamask/package.json
+++ b/examples/omni-lock-metamask/package.json
@@ -10,7 +10,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@ckb-lumos/lumos": "0.19.0-beta.0",
+    "@ckb-lumos/lumos": "0.19.0-alpha.3",
     "@types/react": "^17.0.34",
     "@types/react-dom": "^17.0.11",
     "react": "^17.0.2",
--- a/examples/omni-lock-metamask/package.json
+++ b/examples/omni-lock-metamask/package.json
@@ -10,7 +10,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@ckb-lumos/lumos": "0.19.0-alpha.3",
+    "@ckb-lumos/lumos": "0.19.0-beta.0",
     "@types/react": "^17.0.34",
     "@types/react-dom": "^17.0.11",
     "react": "^17.0.2",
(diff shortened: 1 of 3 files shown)


examples/omni-lock-metamask/package.json
```diff
@@ -10,7 +10,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@ckb-lumos/lumos": "0.19.0-alpha.3",
+    "@ckb-lumos/lumos": "0.19.0-beta.0",
```
homura:
```suggestion
    "@ckb-lumos/lumos": "0.19.0-alpha.3",
```

I found the beta branch is not merged into develop yet, so maybe we don't need to change the dep version now

https://github.com/ckb-js/lumos/blob/next%2Fv0.19.0-beta.0/examples/omni-lock-metamask/package.json

IronLu233: But in `0.19.0-alpha3`, the module do not export `omni` module.

https://github.com/ckb-js/lumos/blob/v0.19.0-alpha.3/packages/common-scripts/src/index.ts#L13

homura: bumped develop to beta.0, we can rm this line, this line would make yarn install lumos from npm instead of using the workspace's lumos

homura (approved):
LGTM

PR #452, opened 2022-10-27
docs: guide for shimming NodeJS API

### LINKED ISSUE #448: [Bug] `@ckb-lumos` packed it's dependencies source code and cause multiple module instance when import from diffierent entries

I found there are two `config-manager` when I run the code.
``` typescript
import { config } from '@ckb-lumos/lumos'
import * as config2 from '@ckb-lumos/config-manager';

const AGGRON4 =config.predefined.AGGRON4;
// set the config in '@ckb-lumos/lumos';
config.initializeConfig(AGGRON4);

// get the config from config-manager package
config2.getConfig(); // return LINA, not AGGRON4
config2.getConfig() !== config.getConfig() // they are not same
```

The package named `@ckb-lumos/lumos` should not include other packages' source code(eg. `@ckb-lumos/config-manager`).

But there are a special case, that we need include the dependencies source code. it's when we want bundle a browser package(use iife, and user call lumos API via `window.lumos`); we can provide a special package for browser.

But in most other case. we don't pack the dependencies source code into `@ckb-lumos/lumos`, just let the bundler, like Webpack, to resolve the package dependencies, and other packages into one.

**Describe alternatives you've considered**

1. Always import the API in `@ckb-lumos/lumos`; Only import it from ”`@ckb-lumos/others`“;'
2. Never import API from `@ckb-lumos/lumos`; Only import it from `@ckb-lumos/others`;'

**Additional context**

image
There a funcion named `getConfig` in `@ckb-lumos/lumos` bundle.

IronLu233:
# Description

In future version, we do not provide NodeJS API polyfills. and we should tell our user how to config their build toolchain(Mostly Webpack or Vite).

This PR add a new secion named "Polyfill NodeJS API", and provide these build tool guides
- Webpack
- Create React App
- Vite

fixed: #448 

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] Document change only
- [x] All document code runs well on local

### DIFF
--- a/website/docs/01-run-lumos-in-the-browser.md
+++ b/website/docs/01-run-lumos-in-the-browser.md
@@ -9,13 +9,14 @@ sidebar_position: 2
 
 Lumos was originally run on NodeJS only. To run on browser, we replaced native indexer with ckb-indexer, added BI which is a big number library, and a series of other upgrades.
 
+Also, we need polyfill the NodeJS API(such as `crypto`), and change your build toolchain config to support.
+
 The following example of getting the balance will show you how to use lumos in your web project.
 
 ```shell
 npm install @ckb-lumos/lumos
 # yarn add @ckb-lumos/lumos
 ```
-
 ```ts
 import { Script, Indexer, BI } from "@ckb-lumos/lumos";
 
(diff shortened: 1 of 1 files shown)


homura (approved):
LGTM

PR #456, opened 2022-11-08
fix(website): fix cannot set properties of undefined

### LINKED ISSUE #455: The Address Conversion tool is broken.
The [address conversion tool](https://lumos-website.vercel.app/tools/address-conversion) is no longer working. A Javascript error is received on page load.

IronLu233:
# Description

Fixes #455

As the issue say, are an error on our website, when we use the address tool.
image

The root cause is about babel transform runtime module.

Docusaurus's build system is based on Webpack and Bebel. When we use the newer Javascript feature, the Babel's preset named `preset-env` will transform it. and the transform work is finished by `transform-runtime` plugin. `preset-env` will read current browserslist config, and then tell the `transform-runtime-plugin` which Js syntax should transform.

Some language features' transform need helper function, so the transform-runtime plugin will inject extra import/require statement for it.

```
var helper = require("babel-helpers");
```

the subpackage of lumos is built with commonjs, but the babel inject es module import statement for it.

The injected source file seems like
``` js
import  classPrivateFieldLooseBase from 'babel-runtime-helpers/esm/classPrivateFieldLooseBase'

exports.foo = () => {  }
```

So the Webpack regard the injected source file as ES module. the import statement be transformed, and expors.xxx remains.

`exports` is not in browser global variable, so it put the error in console.

image

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] use the origin test

### DIFF
--- a/website/babel.config.js
+++ b/website/babel.config.js
@@ -1,3 +1,13 @@
+const babelPresets = require('@docusaurus/core/lib/babel/preset').default;
+
 module.exports = {
-  presets: [require.resolve('@docusaurus/core/lib/babel/preset')],
-};
+  presets: [(api) => {
+    const origin = babelPresets(api);
+    const transformRuntime = origin.plugins.find(plugin => plugin[0] === require.resolve('@babel/plugin-transform-runtime'))
+    if (transformRuntime) {
+      transformRuntime[1].helpers = false
+    }
+    return origin
+  }],
+
+}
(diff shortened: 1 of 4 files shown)


homura:
image

The `code hash` and `hash type` is empty, can you please fix it in this PR?
We should [migrate](https://lumos-website.vercel.app/migrations/migrate-to-v0.19#step-1---codemod) the website to lumos@0.19

https://github.com/ckb-js/lumos/blob/ce18d6d2b15793e61e0e35072aa65ebe4aea5b55/website/src/components/address-conversion/AddressToScript.tsx#L52-L53

IronLu233:
image
image
I migrated it.

homura (approved):
LGTM

PR #445, opened 2022-10-13
feat(ckb-indexer): integrate merged rpc

PainterPuppets:
# Description

- [x] update @ckb-lumos/ckb-indexer to support merged indexer

Resolved: https://github.com/ckb-js/lumos/issues/416

This is a no breaking change PR

With the [merged indexer](https://github.com/nervosnetwork/ckb/pull/3609), we no longer need to depend on a separate ckb-indexer

```ts
import { Indexer } from '@ckb-lumos/lumos';

// before
// const indexer = new Indexer(
//   "https://testnet.ckb.dev/indexer", 
//   "https://testnet.ckb.dev/rpc",
// );

// after
const indexer = new Indexer("https://testnet.ckb.dev/rpc");
```

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactor (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation or Website
- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] pass e2e test

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules -->

PR #458, opened 2022-11-09
refactor(examples): modern examples with codec and merged indexer

IronLu233:
# Description

This PR do two things:
- merge example Indexer class into one parameter, please see: https://github.com/ckb-js/lumos/pull/445
- remove generated molecule code, use `@ckb-lumos/codec` instead

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- Use origin test
- All examples be tested in local


packages/ckb-indexer/src/indexer.ts
```diff
@@ -51,6 +51,9 @@ export class CkbIndexer implements CellProvider, TerminableCellFetcher {
   medianTimeEmitters: EventEmitter[] = [];
   emitters: IndexerEmitter[] = [];
   isSubscribeRunning = false;
+
+  constructor(ckbRpcUrl: string);
```
IronLu233, line 55: it says in https://github.com/ckb-js/lumos/pull/445
``` ts
import { Indexer } from '@ckb-lumos/lumos';

// before
// const indexer = new Indexer(
//   "https://testnet.ckb.dev/indexer", 
//   "https://testnet.ckb.dev/rpc",
// );

// after
const indexer = new Indexer("https://testnet.ckb.dev/rpc");
```

So when it have one parameter, the first parameter means `ckbRpcUrl`, not `ckbIndexerUrl`

examples/omni-lock-metamask/index.tsx
```diff
@@ -49,7 +49,10 @@ export function App() {
 
     transfer({ amount: transferAmount, from: omniAddr, to: transferAddr })
       .then(setTxHash)
-      .catch((e) => alert(e.message || JSON.stringify(e)))
+      .catch((e) => {
+        console.log(e);
```
PainterPuppets, line 53: remove console

PainterPuppets (approved):
LGTM

examples/omni-lock-metamask/index.tsx
IronLu233: I think it's necessary. Because the user may change the example, and put it to console takes more help.

homura (approved):
LGTM

PR #466, opened 2022-11-29
fix(examples): fix missing node polyfill

IronLu233:
# Description

when upgrade to 0.20.0, we need update the parcel dependencies for NodeJs shim.

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] All examples are tested in build

**Nami wallet example**
image

**Secp256k1**
image

### DIFF
--- a/examples/cardano-lock-namiwallet/package.json
+++ b/examples/cardano-lock-namiwallet/package.json
@@ -17,7 +17,7 @@
     "@emurgo/cardano-serialization-lib-asmjs": "^10.0.4",
     "@types/react": "^18.0.25",
     "@types/react-dom": "^18.0.9",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0"
+    "react": "18.2.0",
+    "react-dom": "18.2.0"
   }
 }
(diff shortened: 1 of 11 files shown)


homura:
image


it seems to cause another problem, React cannot be access after launching parcel

examples/cardano-lock-namiwallet/package.json
```diff
@@ -14,10 +14,6 @@
   "dependencies": {
     "@ckb-lumos/lumos": "0.20.0-alpha.0",
     "@emurgo/cardano-message-signing-asmjs": "^1.0.1",
-    "@emurgo/cardano-serialization-lib-asmjs": "^10.0.4",
-    "@types/react": "^17.0.34",
-    "@types/react-dom": "^17.0.11",
-    "react": "^17.0.2",
```
homura, line 20: cannot remove react, which will make examples not work in Codesandbox, recommend to use yarn resolutions + fixed dependency version

homura:
To avoid errors caused by multiple react versions, we should lock the version of react in examples to ensure that examples are using the same version of react

```diff
- "react": "^18.2.0"
+ "react": "18.2.0"
```

homura (approved):
LGTM

PR #475, opened 2023-01-03
feat(ckb-indexer): multiple collector query options

### LINKED ISSUE #470: Collecting cells from more than one script

The programming model of CKB is UTxO, for UTxO wallets, they often derive many keys, which often correspond to different lock scripts, that is, the wallets often need to collect cells with more than one lock. But the current `CellCollector` only supports a single script filter, if lumos support for multiple scripts, it will be convenient for collecting cells with multiple locks

Maybe we can enhance the `CellCollector` to archive this

```ts
const collector = new CellCollector(rpc, [
  { lock: lock1, ... },
  { lock: lock2, ... },
  { lock: lockn, ... },
])

for await (const cell of collector.collect()) {
  ...
}
```

simply says, we can enhance the CellCollector's constructor second param, change it to `CKBIndexerQueryOptions | CKBIndexerQueryOptions[]` https://github.com/ckb-js/lumos/blob/437b641279317a70abc5dd4d0f4790455c13aa61/packages/ckb-indexer/src/collector.ts#L25-L30

IronLu233:
# Description
let class `CKBIndexer` support mutiple queries. With one time invoke `collector.collect()`, all of cells fit the queries will be returned. and they are not duplicate.
```ts
const collector = new CellCollector(rpc, [
  { lock: lock1, ... },
  { lock: lock2, ... },
  { lock: lockn, ... },
])

for await (const cell of collector.collect()) {
  ...
}
```

Resolve #470 

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] Should this feature has no breaking change, all previous unit test and E2E test should pass.
- [x] Add a new unit test for this features. 

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules -->

### DIFF
--- a/packages/ckb-indexer/src/collector.ts
+++ b/packages/ckb-indexer/src/collector.ts
@@ -29,8 +29,6 @@ export class CKBCellCollector implements BaseCellCollector {
     queries: CKBIndexerQueryOptions | CKBIndexerQueryOptions[],
     public otherQueryOptions?: OtherQueryOptions
   ) {
-    this.queries = Array.isArray(queries) ? queries : [queries];
-
     const defaultQuery: CKBIndexerQueryOptions = {
       lock: undefined,
       type: undefined,
@@ -44,7 +42,9 @@ export class CKBCellCollector implements BaseCellCollector {
       outputCapacityRange: undefined,
       bufferSize: undefined,
     };
-    this.queries = { ...defaultQuery, ...this.queries };
+    this.queries = (Array.isArray(queries) ? queries : [queries]).map(
+      (query) => ({ ...defaultQuery, ...query })
+    );
     this.validateQueryOption(this.queries);
     this.convertQueryOptionToSearchKey();
   }
@@ -211,24 +211,37 @@ export class CKBCellCollector implements BaseCellCollector {
   }
 
   async count(): Promise<number> {
-    let lastCursor: undefined | string = undefined;
-    const getCellWithCursor = async (
-      query: CKBIndexerQueryOptions
-    ): Promise<Cell[]> => {
-      const result: GetCellsResults = await this.getLiveCell(query, lastCursor);
-      lastCursor = result.lastCursor;
-      return result.objects;
-    };
     let counter = 0;
+    const visitedCell = new Set<string>();
 
-    for (const queryIndex of this.queries.keys()) {
-      const query = this.queries[queryIndex];
+    for (const query of this.queries) {
+      let lastCursor: undefined | string = undefined;
+      const getCellWithCursor = async (
+        query: CKBIndexerQueryOptions
+      ): Promise<Cell[]> => {
+        const result: GetCellsResults = await this.getLiveCell(
+          query,
+          lastCursor
+        );
+        lastCursor = result.lastCursor;
+        return result.objects;
+      };
+
       let cells: Cell[] = await getCellWithCursor(query);
       let buffer: Promise<Cell[]> = getCellWithCursor(query);
       let index = 0;
       let skippedCount = 0;
       while (true) {
-        if (!this.shouldSkipped(cells[index], query, skippedCount)) {
+        const cell = cells[index];
+        const key = `${cell.outPoint?.txHash}-${cell.outPoint?.index}`;
+        if (visitedCell.has(key)) {
+          index++;
+          continue;
+        } else {
+          visitedCell.add(key);
+        }
+
+        if (!this.shouldSkipped(cell, query, skippedCount)) {
           counter += 1;
         } else {
           skippedCount++;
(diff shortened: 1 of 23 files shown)


packages/ckb-indexer/src/collector.ts
```diff
@@ -41,197 +42,181 @@ export class CKBCellCollector implements BaseCellCollector {
...
     this.validateQueryOption(this.queries);
     this.convertQueryOptionToSearchKey();
   }
 
-  public validateQueryOption(queries: CKBIndexerQueryOptions): void {
-    if (!queries.lock && (!queries.type || queries.type === "empty")) {
-      throw new Error("Either lock or type script must be provided!");
-    }
+  public validateQueryOption(queries: CKBIndexerQueryOptions[]): void {
+    queries.forEach((query) => {
```
homura: Maybe we don't need to modify this method

```js
this.queries.forEach(this.validateQueryOption.bind(this))
```

this will make the nesting depth less and make the code look cleaner

homura:
@IronLu233 can you please add some unit test for this PR to improve the coverage

packages/ckb-indexer/tests/cell_collector.unit.test.ts
```diff
@@ -83,9 +85,31 @@ test("convertParams# should match scriptLenRange if type is 'empty' and scriptLe
...
+      scriptLenRange: ["0x0", "0xff"],
+      type: "empty",
+      order: "asc",
+    },
+  ]);
+
+  t.deepEqual(
+    cellCollector.queries.flatMap((query) => query.scriptLenRange),
+    ["0x0", "0x1", "0x0", "0xff"]
+  );
```
homura: maybe it is a meaningless case when it is flatted

```ts
t.deepEqual(
  collector.queries.map(query => query.scriptLenRange),
  [['0x0', '0x1'], ['0x0', '0xff']]
)
```

packages/ckb-indexer/tests/cell_collector.unit.test.ts
```diff
@@ -208,3 +232,133 @@ test("validateQueryOption#validate scriptLenRange", (t) => {
...
+    "throw error if lock and query both not provided in queryOption[1]"
+  );
+
+  t.throws(
+    () => {
+      new CellCollector(indexer, [{ lock: lockScript }, { type: "empty" }]);
+    },
+    undefined,
+    "throw error if lock is not provided and type is empty in queryOption[1]"
+  );
```
homura, line 254: why `{ type: "empty" }` will be throw here?

packages/ckb-indexer/tests/collector.e2e.test.ts
```diff
@@ -32,6 +33,34 @@ test("get count correct", async (t) => {
...
+    { lock: type },
+    {
+      lock: {
+        ...type,
+        args: "0x7ae354c3586ea3e7da6f30af80046fbe0cdce2fd",
+      },
+    },
+  ]);
+  const count = await cellCollector.count();
+  t.is(count, 2);
```
homura, line 53:
```ts
const collected: Cell[] = []
for await (let cell of cellCollector.collect()) {
  collected.push(cell)
}

t.is(collected[0].lock.args, '0xa178db16d8228db82911fdb536df1916e761e205')
t.is(collected[1].lock.args, '0x7ae354c3586ea3e7da6f30af80046fbe0cdce2fd')
```

this is a less strict e2e case, the count of cells is 2 is possible that the same lock

IronLu233: I think there are two other case for your concern.

https://github.com/ckb-js/lumos/pull/475/files#diff-61ebc5734333dcf70b862d371fe1b5f2ae917c40c1162d80a2706e9ca1fd017dR121

and

https://github.com/ckb-js/lumos/pull/475/files#diff-61ebc5734333dcf70b862d371fe1b5f2ae917c40c1162d80a2706e9ca1fd017dR96

Here, in this case I just test the count method, not the collect method. although the `count` method rely on `collect`.

packages/ckb-indexer/tests/cell_collector.unit.test.ts
IronLu233: https://github.com/ckb-js/lumos/pull/475/files#diff-f9c531afc5c4264761e548055e5ab3b3b6b33ce8754fdac1fdb196c2e7212155R129

But it is copied from here. I just reuse previous code.

homura: the diff link seems broken, can you give a permalink?

IronLu233: https://github.com/IronLu233/lumos/blob/c49e0bc3b191cc4aeb53487581eee7df2f646f18/packages/ckb-indexer/tests/cell_collector.unit.test.ts#L127

homura: perhaps the user story that can only be explained as "collect lock only cells" does not make sense

homura (approved):

PR #476, opened 2023-01-05
chore(hd): deprecate load/save method

zhangyouxin:
# Description

This PR deprecates load and save method in `@ckb-lumos/hd` package.

# Motivation

`@ckb-lumos/hd` was designed for node environment, to make it more compatible in browser env, we need deprecate load/save method which operate with `fs` module and tell front-end project developer how to use `@ckb-lumos/hd`.

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] Unit test


.gitignore
```diff
@@ -109,6 +109,7 @@ packages/sql-indexer/knexfile.dev.js
 packages/helpers/lib
 packages/common-scripts/lib
 packages/hd/lib
+packages/hd/*.json
```
homura: this causes `package.json` to be ignored, why we should do this

zhangyouxin: Added test case for save/load keystore, the file name is {uuid}.json, maybe I should move the json files created by the test case to another dir.

packages/hd/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
```
homura: the changelog should be auto-generated, we don't need to edit it manually, we can rm it

homura (approved):

PR #483, opened 2023-01-09
feat(hd): browser compatible scrypt

zhangyouxin:
# Description

Since `scryptSync` is not implemented in `crypto-browserify`, we need to use standalone `scrypt-js` to make `@ckb-lumos/hd` compatible in browser enviroment. 

This PR is a trade-off between performance and compatibility because using `scrypt-js` takes about **254%** more time than `crypto.scryptSyc(...)`:

```
crypto.scryptSync x 1.49 ops/sec ±0.50% (53 runs sampled)
scrypt-js x 0.42 ops/sec ±0.18% (51 runs sampled)
```

<details><summary>benchmark.ts</summary>

```ts
import Benchmark from "benchmark";
import crypto from "crypto";
import { syncScrypt } from "scrypt-js";

const pwd = "testpassword";
const salt: Buffer = crypto.randomBytes(32);
const kdfparams = {
  dklen: 32,
  salt: salt.toString("hex"),
  n: 2 ** 18,
  r: 8,
  p: 1,
};
Benchmark.options.minSamples = 50;
const suite = new Benchmark.Suite();

suite
  .add("crypto.scryptSync", () => {
    return crypto.scryptSync(
      pwd,
      salt,
      kdfparams.dklen,
      {
        N: kdfparams.n,
        r: kdfparams.r,
        p: kdfparams.p,
        maxmem: 128 * (kdfparams.n + kdfparams.p + 2) * kdfparams.r,
      }
  )}
  )
  .add("scrypt-js", () => {
    return Buffer.from(
      syncScrypt(
        Buffer.from(pwd),
        salt,
        kdfparams.n,
        kdfparams.r,
        kdfparams.p,
        kdfparams.dklen
      )
    );
  })
  .on("cycle", function (event: any) {
    console.log(String(event.target));
  })
  .run();
```

</details>

And according to actual performance in CI, it takes roughly 4 times of time when using `scrypt-js`, this result may be not that accurate due to multi-thread tasks and other environment tweaks in CI, just for reference:

Takes about 13 seconds using `scrypt-js`: https://github.com/ckb-js/lumos/actions/runs/3879191496/jobs/6616131411#step:7:571
Takes about 3 seconds using `crypto.scryptSync`:  https://github.com/ckb-js/lumos/actions/runs/3880536712/jobs/6618674890#step:7:451

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] added test vector from [web3 secret storage 101](https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition#test-vectors)


homura:
i found this PR's some tests are 5x - 7x slower than before, we can check out these cases, such as `loadFromKeystore`, `derive threshold`, `getNextReceivingPublicKeyInfo`, etc.

packages/hd/src/keystore.ts
```diff
@@ -234,11 +239,15 @@ export default class Keystore {
 
   derivedKey(password: string): Buffer {
     const { kdfparams } = this.crypto;
-    return crypto.scryptSync(
-      password,
-      Buffer.from(kdfparams.salt, "hex"),
-      kdfparams.dklen,
-      Keystore.scryptOptions(kdfparams)
```
homura, line 241: this is a breaking change here, we need to be compatible with the previous `kdfparams`

packages/hd/tests/keystore.test.ts
```diff
@@ -44,6 +44,34 @@ test("load and check password, loads private key", (t) => {
...
+        r: 1,
+        salt: "ab0c7876052600dd703518d6fc3fe8984592145b591fc8fb5c6d43190334ba19",
+      },
+      mac: "2103ac29920d71da29f15d75b4a16dbe95cfd7ff8faea1056c33131d846e3097",
+    },
+    id: "3198bc9c-6672-5ab3-d995-4942343ae5b6",
+    version: 3,
+  };
+  const keystore = Keystore.fromJson(JSON.stringify(json));
+  t.true(keystore.checkPassword("testpassword"));
```
homura, line 84: we can decrypt the json here to check if the decrypted secret is same with original secret

packages/hd/src/keystore.ts
zhangyouxin: The method signature stays the same: `derivedKey: (password: string) => Buffer`

homura: I think I misunderstood, this is not a breaking change, just ignore this comment

packages/hd/tests/keystore.test.ts
zhangyouxin: sure, added assertion for `decrypt` and `derivekey`

homura (approved):

PR #486, opened 2023-02-23
fix: (transaction-manager) some issues of lumos

oriole-pub:
# Description

This PR fixes the following two issues：

Two problems were found when using transaction-manager.

1、An undefined exception occurred in the internal method _checkTransactions of transaction-manager, which made it impossible to clean up createdCells normally. 

#### Because of
TransactionCollector. getTransactionHashes() to throw an undefined exception, because there is an RPC call inside this method, but the returned result is not formatted

``` ts
... /lumos/packages/ckb-indexer/src/transaction_collector.ts

private getTransactionListFromRpc = async (
    indexerTransactionList: IndexerTransactionList
  ) => {
    ///....
    const transactionList: TransactionWithStatus[] = await services
      .requestBatch(this.CKBRpcUrl, getDetailRequestData)
      .then((response: GetTransactionRPCResult[]) => {
        return response.map(
          (item: GetTransactionRPCResult): TransactionWithStatus => {
            if (!this.filterOptions.skipMissing && !item.result) {
              throw new Error(
                `Transaction ${
                  indexerTransactionList.objects[item.id].txHash
                } is missing!`
              );
            }
           // The result was not converted
            return { ...item.result };
          }
        );
      });
    return transactionList;
  };

public getResolvedTransactionRequestPayload(
    unresolvedTransactionList: TransactionWithStatus[],
    indexerTransactionList: IndexerTransactionList
  ): JsonRprRequestBody[] {
    const requestPayload: JsonRprRequestBody[] = [];
    let resolvedTransactionRequestId = 0;
    unresolvedTransactionList.forEach(
      (unresolvedTransaction: TransactionWithStatus, index: number) => {
        const indexerTransaction = indexerTransactionList.objects[index];
        if (indexerTransaction.ioType === "input") {

          //  Exception will occur here
          //  Due to the case problem of field name, the actual field cannot be retrieved, resulting in undefined
          //  unresolvedOutPoint = undefined
          // unresolvedOutPoint.previousOutput = undefined
          // unresolvedOutPoint.previousOutput.txHash = undefined

          const unresolvedOutPoint: OutPoint =
            unresolvedTransaction.transaction.inputs[
              Number(indexerTransaction.ioIndex)
            ].previousOutput;
          requestPayload.push({
            id: resolvedTransactionRequestId++,
            jsonrpc: "2.0",
            method: "get_transaction",
            params: [unresolvedOutPoint.txHash],
          });
        }
      }
    );
    return requestPayload;
  }

```

2、`TransactionManager.collect()` did not filter the spent `createdCell`, resulting in the failure to return the correct `live cell ` when multiple transactions were initiated consecutively in a short period of time.

**This site can reproduce the problem**

If the `get test token` transaction is initiated more than 3 times in a row, an error message will appear

https://godwoken-bridge-testnet-git-fork-gitofjason-fix-b13e86-godwoken.vercel.app/#/v1/deposit/pending
image

### Supplementary 

The `TransactionManager.collector` method adds the `outputDataLenRange` parameter, which is needed to filter cells in some scenarios

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

### DIFF
--- a/packages/ckb-indexer/src/indexer.ts
+++ b/packages/ckb-indexer/src/indexer.ts
@@ -12,7 +12,6 @@ import { requestBatch } from "./services";
 import { CKBCellCollector } from "./collector";
 import { EventEmitter } from "events";
 import {
-  GetTransactionRPCResult,
   CKBIndexerQueryOptions,
   GetCellsResults,
   GetLiveCellsResult,
@@ -34,6 +33,8 @@ import { RPC as CKBIndexerRpc } from "./rpc";
 import { CKBRPC } from "@ckb-lumos/rpc";
 import { validators } from "@ckb-lumos/toolkit";
 import type * as IndexerType from "./indexerType";
+import type * as RPCType from "./rpcType";
+import { toOutput } from "@ckb-lumos/rpc/lib/resultFormatter";
 
 const DefaultTerminator: Terminator = () => {
   return { stop: false, push: true };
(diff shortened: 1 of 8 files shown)


zhangyouxin:
@GitOfJason This fix looks good to me, and it will look great if a test case is added to cover it, and to make it pass the codecov check BTW

oriole-pub:
> @GitOfJason This fix looks good to me, and it will look great if a test case is added to cover it, and to make it pass the codecov check BTW

Yes, this is exactly what i need to do

oriole-pub:
@zhangyouxin  hi~  I think the unit test has covered the part I modified, but why codecov still fails?

oriole-pub:
@zhangyouxin  Are there any further requests for this PR?

zhangyouxin:
There are some annotations indicating line not covered by test, such as the annotation below [this line](https://github.com/ckb-js/lumos/pull/486/files#diff-76de6be682640bbe843589fcdde866489bb7d2f99b11c7c2887b9825dd4762d0R298), we may follow this hint and cover it by test so that the patch coverage will hit the target.

homura:
We completely rewrote the [transaction manager](https://github.com/ckb-js/lumos/blob/develop/packages/transaction-manager/README.md), which now works with NodeJS and the browser, and also supports custom caching so that page refreshes can also cache pending cells,  if you want to experience it, you can install `@ckb-lumos/transaction-manager@0.20.0- alpha.3`

PR #492, opened 2023-03-03
fix: multiple changes to the transaction-manage

oriole-pub:
# Description
This merge request is #486 split, and makes the following changes:

- Transaction-manager will now check whether the createdCell has been spent.
- Collector will now add an outputDataLenRange parameter.
- TransactionCollector will now include the missing parameter CKBRpcUrl.
- Adjustment to the method name in index.d.ts.

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

### DIFF
--- a/packages/transaction-manager/index.d.ts
+++ b/packages/transaction-manager/index.d.ts
@@ -6,6 +6,7 @@ import {
   Indexer,
   CellCollector,
 } from "@ckb-lumos/base";
+import { HexadecimalRange } from "@ckb-lumos/ckb-indexer/lib/type";
 
 declare class TransactionManager {
   constructor(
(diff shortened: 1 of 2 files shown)


homura:
Thanks for your PR, we completely rewrote the [transaction manager](https://github.com/ckb-js/lumos/blob/develop/packages/transaction-manager/README.md?rgh-link-date=2023-06-08T13%3A43%3A50Z) since there were too many issues in transaction manager, which now works with NodeJS and the browser, and also supports custom caching so that page refreshes can also cache pending cells, if you want to experience it, you can install @ckb-lumos/transaction-manager@0.20.0- alpha.3

Issue #501, opened 2023-04-25
Support Chained Transactions


zhangyouxin:
## Motivation

https://talk.nervos.org/t/non-blocking-chained-transaction-and-its-applications-in-ckb/6649

In order to provide a more smooth experience for developer/dapp user to send transactions in sequence, we need to collect freshly created pending cells created by recent transactioins, although they are still in memPool.

## What can we do to support it

To achieve the purpose of this issue described in motivation, we need to implement:

- ckb-indexer filter protocol, see the `search_key`, `limit`, `order` and `after` in [get_cells rpc](https://github.com/nervosnetwork/ckb/blob/develop/rpc/README.md#method-get_cells)
- transaction manager to manage new created transaction outputs and inputs

What we need to do:

- [ ] implement ckb-indexer filter protocol in @ckb-lumos/ckb-indexer
- [ ] refactor the current cell collector to use the filter
- [ ] the @ckb-lumos/transaction-manager is too old, create a new transaction-manager with reasonable API

homura:
update: after the discussion, we found that we can update the existing [`TransactionManager`](https://github.com/ckb-js/lumos/blob/develop/packages/transaction-manager/index.d.ts) to handle the pending cells

```diff
+ interface CkbIndexerCellProvider {
+    collector(query: CkbIndexerQueryOptions): { collect(): AsyncIterator<Cell> }
+ }

declare class TransactionManager {
  constructor(
-   indexer: Indexer,
+   // when a string is provided, it will be treated as a CKB RPC
+   indexer: CkbIndexerCellProvider | string
    options?: {
      logger?: logger.Logger;
      pollIntervalSeconds?: number;
      storage?: AsyncStorage;
    }
  );

  start(): void;

  stop(): void;

  sendTransaction(tx: Transaction): Promise<Hash>;

  /**
   * If set `usePendingOutputs` to false, will not use pending outputs, default to true.
   *
   * @param queryOptions
   * @param options
   */
  collector(
-    queryOptions?: QueryOptions,
+    queryOptions?: CkbIndexerQueryOptions,
    options?: { usePendingOutputs?: boolean }
  ): CellCollector;
}


```

zhangyouxin:
It's reasonable to change `indexer` to `CkbIndexerCellProvider | string`, also we need to pass in `rpcUrl` in constructor to make sure `TransactionManger` is able to:

- send transactions to rpc endpoint
- update stored transaction status


```diff
  constructor(
[...]

PR #502, opened 2023-04-27
feat: indexer cells filter

zhangyouxin:
# Description

This pr implements an cells filter protocol identical to ckb-indexer filter: https://github.com/nervosnetwork/ckb/blob/develop/rpc/README.md#method-get_cells

This pr resolve a task in #501 

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] Unit Tests


packages/ckb-indexer/src/ckbIndexerFilter.ts
```diff
@@ -0,0 +1,230 @@
...
+  script: Script | undefined,
+  scriptLenRange: HexadecimalRange
+): boolean {
+  const scriptLen = script
+    ? BI.from(
+        bytes.bytify(script.codeHash).length +
+          bytes.bytify(script.args).length +
+          1 /* hashType length is 1 */
+      )
+    : BI.from(0);
```
homura:
```ts
blockchain.Script.pack(script)
```

is fine

packages/ckb-indexer/src/ckbIndexerFilter.ts
```diff
@@ -0,0 +1,341 @@
...
+}
+
+const unWrapScriptWrapper = (inputScript: ScriptWrapper | Script): Script => {
+  if (instanceOfScriptWrapper(inputScript)) {
+    return inputScript.script;
+  }
+  return inputScript;
+};
+
+const unWrapDataWrapper = (input: DataWithSearchMode | string): string => {
```
homura: `unwrap` can be a word

packages/ckb-indexer/src/ckbIndexerFilter.ts
```diff
@@ -0,0 +1,341 @@
...
+        bytes.equal(bytes.bytify(cell.data), bytes.bytify(dataSearch.data))
+      );
+    } else if (
+      instanceOfDataWithSearchMode(options.data) &&
+      options.data.searchMode === "prefix"
+    ) {
+      const dataSearch = options.data as DataWithSearchMode;
+      filteredCells = filteredCells.filter(
+        (cell) =>
+          Buffer.from(bytes.bytify(cell.data)).indexOf(
```
homura: We should avoid using the NodeJS API as possible as we can

```ts
const expectPrefix = bytes.bytify(dataSearch.data);
const actualPrefix = cell.data.slice(0, expectPrefix.length);
return bytes.equal(expectPrefix, actualPrefix);
```

packages/ckb-indexer/src/ckbIndexerFilter.ts
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  if (options.skip) {
+    filteredCells = filteredCells.slice(options.skip);
+  }
+
+  return filteredCells;
+}
+
+function filterBy(cell: Cell, searchKey: SearchKey): boolean {
```
homura: Mark it as `@internal` or remove it from export list, because we should avoid functions being used incorrectly by developers

packages/ckb-indexer/src/ckbIndexerFilter.ts
```diff
@@ -0,0 +1,341 @@
...
+        return false;
+      }
+    } else {
+      if (!checkScriptWithPrefixMode(cellOutput.type, script)) {
+        return false;
+      }
+    }
+  }
+
+  // Filter is always in prefix mode
```
homura:
```ts
// the "exact" mode works only on "SearchKey.script", 
// not on "SearchKey.filter.script"
// the "SearchKey.filter.script" is always in prefix mode
```

packages/ckb-indexer/src/ckbIndexerFilter.ts
```diff
@@ -0,0 +1,341 @@
...
+  if (filter) {
```
homura:
```ts
if(!filter) return true;
```

early return makes the code more readable

packages/ckb-indexer/src/ckbIndexerFilter.ts
```diff
@@ -0,0 +1,341 @@
...
+      if (dataLen.lt(fromDataLen) || dataLen.gte(toDataLen)) {
+        return false;
+      }
+    }
+  }
+
+  return true;
+}
+
+function isCellAfterCursor(payload: { cell: Cell; cursor: string }): boolean {
```
homura: maybe it is unnecessary in client

packages/ckb-indexer/src/collector.ts
```diff
@@ -199,16 +204,39 @@ export class CKBCellCollector implements BaseCellCollector {
...
     if (
+      query.argsLen !== undefined &&
       query.argsLen !== -1 &&
       query.argsLen !== "any" &&
       getHexStringBytes(cell.cellOutput.lock.args) !== query.argsLen
     ) {
       return true;
     }
+
+    if (!!query.data && query.data !== "any") {
```
```ts
if (query.data && query.data === 'any') return false 
```

packages/ckb-indexer/src/indexerCursor.ts
homura, line 1: maybe this file is unnecessary

packages/ckb-indexer/tests/cell_collector.unit.test.ts
```diff
@@ -295,8 +295,21 @@ test("shouldSkip#shouldSkip should works well", (t) => {
...
+      { data: "0x1234" }
+    )
+  );
+  t.true(cellCollector.shouldSkipped({ data: "0x5678" }, { data: "0x1234" }));
+  t.true(
+    cellCollector.shouldSkipped(
+      { data: { data: "0x5678", searchMode: "exact" } },
+      { data: "0x1234" }
+    )
+  );
```
homura: please add some test for `prefix`

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: `blockchain.Script` is a `table`, so the packed size will be more than raw data, there are some length info in the packed data.

homura: got it, maybe we can refactor to 

```ts
bytes.concat(script.codeHash, args).length + 1 // the last + 1 is for hash_type
```
to make it simpler

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: Is it ok to just not re-export it in `index.ts`, not exporting this function may lead to inconvenience in testing

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: updated

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: updated

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: removed all `Buffer`

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: updated

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: updated

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: removed

packages/ckb-indexer/src/collector.ts
zhangyouxin: updated

packages/ckb-indexer/src/indexerCursor.ts
zhangyouxin: removed

packages/ckb-indexer/tests/cell_collector.unit.test.ts
zhangyouxin: added

homura:
https://github.com/ckb-js/lumos/blob/0c5ea9feea5acf0e3dfa48b58d6a2a9de2b0fa90/packages/ckb-indexer/src/collector.ts#L55-L172 

please refactor to the new convert to improve the test coverage

zhangyouxin:
> https://github.com/ckb-js/lumos/blob/0c5ea9feea5acf0e3dfa48b58d6a2a9de2b0fa90/packages/ckb-indexer/src/collector.ts#L55-L172
> 
> please refactor to the new convert to improve the test coverage

The code you referred does not convert `QueryOptions` to `SearchKey`, the original convert function is here: https://github.com/ckb-js/lumos/blob/develop/packages/ckb-indexer/src/services.ts#L18 , which is not compatible with new convert funciton, its parameter type is `CKBIndexerQueryOptions`

zhangyouxin:
Do we need to put `CKBIndexerQueryOptions` in `@ckb-lumos/base`, and enhance new convertor to accept `CKBIndexerQueryOptions` as input parameter type?

homura:
I guess it would be better to make progressive changes, so just keep implementing it in `@ckb-lumos/ckb-indexer`. If we base on the idea of progressive changes, can we

1. extend the matcher so that it supports `CkbIndexOptions`
```ts
declare function isMatchQueryOption(cell: Cell, options:QueryOption); 

function isMatchCkbIndexerQueryOption(cell: Cell, options: CkbIndexerQueryOptions) {
  if (!filterByQueryOptions(options)) return false;
  // continue to implement another filter 
}
```
2. only a match function to support `CkbIndexerQueryOptions` in one step

```ts
declare functoin isMatchCkbIndexerQueryOption(cell, options: CkbIndexerQueryOptions): boolean;
```

But I haven't considered which one is more appropriate, or is there a more appropriate way?

---

More importantly, it occurred to me that we implemented the `match(or filter)` to provide a more reliable `TransactionManager` or enhance the current `TransactionManager`, in order to keep the behavior in the dApp as consistent as possible with the lumos's `CkbIndexer`, and to filter pending cells intuitively. 

Therefore, I think we can consider how to enhance the `TranasctionManager` or design a new `TranasctionManager` before deciding how to implement the match

zhangyouxin:
The interface of `NewTransactionManager` has been discussd already:

```ts
const manager = new TransactionManager("https://testnet.ckb.dev");

const cellCollector = manager.collector({
  lock: {
    codeHassh: "0x...",
    hashType: "type",
    args: "0x",
  },
});

for await (const cell of cellCollector.collect()) {
  console.log(cell);
}
```

so the cells will be filtered when cellCollector collects cells, maybe option 2 is ok

packages/ckb-indexer/src/ckbIndexerFilter.ts
homura: `@internal` will only just strip the type declaration from compiled `.d.ts`, it does not affect the test

```ts
/**
 * @internal
 */
functoin filterBy() {

```

packages/ckb-indexer/src/ckbIndexerFilter.ts
```diff
@@ -0,0 +1,352 @@
...
+    return false;
+  }
+
+  const expectCodeHashPrefix = bytes.bytify(filterScript.codeHash);
+  const actualCodeHashPrefix = bytes
+    .bytify(script.codeHash)
+    .slice(0, expectCodeHashPrefix.length);
+  if (!bytes.equal(expectCodeHashPrefix, actualCodeHashPrefix)) {
+    return false;
+  }
```
homura: the `codeHash` can only be exact match, since the `codeHash` is always 32 bytes

packages/ckb-indexer/src/ckbIndexerFilter.ts
```diff
@@ -0,0 +1,352 @@
...
+    filteredCells = filteredCells.slice(options.skip);
+  }
+
+  return filteredCells;
+}
+
+/**
+ * @internal
+ */
+function filterBy(cell: Cell, searchKey: SearchKey): boolean {
```
homura: `filterBySearchKey`

packages/ckb-indexer/tests/ckbIndexerFilters.unit.test.ts
```diff
@@ -0,0 +1,267 @@
...
+  t.false(
+    filterBy(mockCell, {
+      scriptType: "lock",
+      script: script1,
+      filter: { outputDataLenRange: ["0x10", "0x11"] },
+    })
+  );
+});
+
+test("should filterByQueryOptions return expected cells", async (t) => {
```
homura: can you please separate these tests like this

```ts
describe("Match with QueryOptions.lock or QueryOptions.type", () => {
  test("script defaults to exact mode")
  test("script.args with exact mode")
})
descripbe("Match with ScriptWrapper", () => {
  test("script only")
  test("with searchMode, with exact mode")
  test("with searchMode, with prefix mode")
  test("with ioType")
  test("with argsLen")
})
```

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: updated

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: updated

packages/ckb-indexer/tests/ckbIndexerFilters.unit.test.ts
zhangyouxin: updated

packages/ckb-indexer/src/ckbIndexerFilter.ts
zhangyouxin: updated

packages/ckb-indexer/src/ckbIndexerFilter.ts
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  if (queryType === "empty") {
+    searchKey.filter.scriptLenRange = ["0x0", "0x1"];
+  }
+
+  return searchKey;
+}
+
+function filterByQueryOptions(
```
homura: I think we should realize that the `SearchKey` is for search/filter/match, and the `order` and `limit` is not for searching

image

This method is used to filter the cells that match the conditions, it should not be mixed with the sorting limit and other operations. We can think of how we would use it 

```ts
async function *collectWithPendingCells(): AsyncGenerator<Cell> {
  for await (let cell of indexer.get(convertToSearchKey(options, cursor)) {
    if (isMatchQueryOption(cell, options)) yield cell; 
  }

  for (let cell of pendingCells) {
    if(isMatchQueryOption(cell, options)) yield cell
  }
}
```

---

update: so we need a `LumosSearchKey` to `Pick` some fields from `CkbIndexerQueryOption`, and comment why we need the `LumosSearchKey` for `filterByQueryOptions`, the `LumosSearchKey` only used to match cell locally

zhangyouxin: updated

homura:
@zhangyouxin Can you revert the merge of https://github.com/ckb-js/lumos/pull/503, because the client side filter and `TransactionManager` can be two features. Only the #502 will be appended to changelog if we put #502 and #503 in one PR. I suggest we can merge client side filter, and open another `TransactoinManager` PR to target to develop

zhangyouxin:
> @zhangyouxin Can you revert the merge of #503, because the client side filter and `TransactionManager` can be two features. Only the #502 will be appended to changelog if we put #502 and #503 in one PR. I suggest we can merge client side filter, and open another `TransactoinManager` PR to target to develop

OK, reseted, #503 is based on #502, so we can merge #502 first

homura (approved):

PR #503, opened 2023-05-04
feat(transaction-manager): pending transaction manager

### LINKED ISSUE #501: Support Chained Transactions

zhangyouxin:
# Description

To enable chained transaction off-chain, it is necessary to store just created `Cells` and just spent `Cells` created by just sent `Transactions`, `TransactiomManager` provides a simple Interface for developer to manage just sent `Transactions`, and the `CellCollector` created by `TransactionManager` is coherent with lumos `@ckb-lumos/ckb-indexer`， just return more cells which are still pending in memory pool.

```ts
interface TransactionManager {
  sendTransaction(tx: Transaction): Promisable<string>;
  collector(queryOptions: CKBIndexerQueryOptions): Promisable<CellCollector>;
}
```

## How does it work?

`PendingTransactionsManager` stores the transactions sent by `sendTransaction` method above to resolve `pendingCells` and `spentCells`, when you collect on-chain cell that is in pending cells, the corresponding pending transaction will be removed.

**!!! There should be only one `PendingTransactionsManager` instance in your app.**

## Usage

```ts
  const manager = new RPCTransactionManager({
    rpcUrl: "https://testnet.ckb.dev",
  });

  const cellCollector = await manager.collector(queryOptions);
  for await (const cell of cellCollector.collect()) {
    // do something with the cell
  }
```

The manager also supports persisting the pending state in a custom storage:

```ts
  // store transactions in browser localstorage 
  const manager = new RPCTransactionManager({
    rpcUrl: "https://testnet.ckb.dev",
    storage: myCustomStorage
  });
```

Please refer to: [README.md](https://github.com/ckb-js/lumos/blob/1811bb2e667b0c3c649ddafb3df331923a458c08/packages/pending-transactions-manager/README.md)

Fixes #501 

## Type of change

- [ ] Example
- [ ] Other

# How Has This Been Tested?

- [x] Unit Test
- [x] An Example is provided to validate it works on testnet https://github.com/ckb-js/lumos/blob/f372922db65e23de01914a4461acdb37d8893bee/packages/transaction-manager/examples/chained-transfer-example/README.md


homura:
It is better to work on `TransactoinManager` instead adding another new API

.DS_Store
homura, line 1: unnecessary, we should add append `.DS_Store` to `.gitignore`

packages/transaction-manager/src/index.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  constructor(payload: {
+    providers: ServiceEndPoint | ServiceProviders;
+    options?: Options;
+  }) {
+    if (isServiceEndPoint(payload.providers)) {
+      const { rpcUrl } = payload.providers;
+
+      this.transactionSender = new RPC(rpcUrl);
+      this.cellCollectorProvider = new Indexer(rpcUrl).collector;
```
homura: May leave a bug here, since the `instance.method` usually needs a `this` of context

image

maybe we can use a `LumosCellIndexer` to instead

```ts
interface LumosCellIndexer {
  collector(options: CkbIndexerOptions): AsyncGenerator<Cell>;
}

this.indexer.collector({...})

```

packages/transaction-manager/src/index.ts
```diff
@@ -0,0 +1,222 @@
...
+    this.txStorage =
+      payload.options?.txStorage || createInMemoryPendingTransactionStorage();
+  }
+
+  async sendTransaction(
+    tx: Transaction,
+    validator: OutputsValidator = "passthrough"
+  ): Promise<string> {
+    const spentCellOutpoints = await this.txStorage.getSpentCellOutpoints();
+    tx.inputs.forEach((input) => {
```
homura: the check is unnecessary here, maybe the user find that a transaction has sent failed and try to send again

packages/transaction-manager/src/PendingTransactionStorage.ts
```diff
@@ -0,0 +1,107 @@
...
+
+type TransactionWithHash = Required<Transaction>;
+
+interface ManagerStorageScheme {
+  transactions: TransactionWithHash[];
+  spentCellOutpoints: OutPoint[];
+  pendingCells: PendingCell[];
+}
+
+export interface TransactionStorage {
```
homura: we should mark as internal here

packages/transaction-manager/src/index.ts
```diff
@@ -0,0 +1,222 @@
...
+      usePendingCells,
+      liveCellCollector,
+      removePendingCell,
+    } = payload;
+
+    this.order = payload.order || "asc";
+    this.spentCells = spentCells;
+    this.filteredPendingCells =
+      payload.order === "desc"
+        ? filteredPendingCells.reverse()
```
homura: reverse is a mutable method, it is better to change it to `array.slice().reverse()`

packages/transaction-manager/src/index.ts
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  async removePendingCell(cell: Cell): Promise<boolean> {
+    return await this.txStorage.deleteTransactionByCell(cell);
+  }
+
+  async collector(
+    options: CKBIndexerQueryOptions,
+    usePendingCells = true
+  ): Promise<PendingCellCollector> {
```
homura: Can you change the return type to `Promise<CellCollector>` or `CkbIndexerCellCollector`? Since the `PendingCellCollector` has many additional methods, and users may be dependent on these APIs, which makes it difficult to make changes in the futuer, as we may have to consider compatibility issues

packages/transaction-manager/src/index.ts
```diff
@@ -0,0 +1,222 @@
...
+      usePendingCells,
+      liveCellCollector,
+      order: options.order,
+      removePendingCell: this.removePendingCell,
+    });
+  }
+}
+
+class PendingCellCollector implements CellCollector {
+  spentCells: OutPoint[];
```
homura: whenever possible, mark members as private

packages/transaction-manager/src/storage.ts
```diff
@@ -0,0 +1,32 @@
+export type Promisable<T> = PromiseLike<T> | T;
+
+/**
+ * key-value storage
+ */
+export interface Storage<Schema> {
+  hasItem<K extends keyof Schema>(key: K): Promisable<boolean>;
+  getItem<K extends keyof Schema>(key: K): Promisable<Schema[K] | undefined>;
+  removeItem<K extends keyof Schema>(key: K): Promisable<boolean>;
```
homura: `void` here to keep the API similar with `localStorage`, which makes user easy to use it, and a low-level API should be simple when possible

```ts
{
  removeItem(key): Promisable<void>
}
```

packages/transaction-manager/src/index.ts
zhangyouxin: updated

packages/transaction-manager/examples/chained-transfer-example/src/index.ts
```diff
@@ -0,0 +1,102 @@
...
+  const txHash2 = await transferViaTxManager();
+  console.log('txHash1:', txHash1);
+  console.log('txHash2:', txHash2);
+}
+
+main();
+
+
+async function transferViaTxManager(): Promise<Hash> {
+  const cells = await collectCells();
```
homura: maybe we can just use `common.transfer` to avoid making the example too complex


```ts
const manager = new TransactionManager();
let txSkeleton = await common.transfer(TransactionSkeleton({ cellProvider: manager }), ...);
// ...
```

packages/transaction-manager/README.md
```diff
@@ -1,24 +1,73 @@
...
+    options: {
+      txStorage:  new PendingTransactionStorage(createBrowserStorage()),
+    },
+  });
+
+  function createBrowserStorage() {
+    const store: Storage = window.localStorage;
+    return {
+      getItem(key) {
+        const value = store.getItem(key) as string | undefined;
```
homura: we should add tips for custom key to avoid conflict keys

## Avoid Storage Conflict

```ts
const CUSTOM_KEY_PREFIX = '__lumos_store__'

getItem(key) {
  const value = store.getItem(CUSTOM_KEY_PREFIX + key)
}
...
```

packages/transaction-manager/README.md
```diff
@@ -1,24 +1,73 @@
...
-
-TransactionManager is a tool for manager uncommitted cells, you can `send_transaction` via this tool and get uncommitted outputs by `collector`.
+# `@ckb-lumos/pending-transactions-manager`
 
 ## Usage
 
-```javascript
-const TransactionManager = require("@ckb-lumos/transaction-manager");
-const { Indexer } = require("@ckb-lumos/ckb-indexer");
+You can create a `PendingTransactionsManager` by passing a `rpcUrl` and an optional `options` object.
```
homura: update the docs

packages/transaction-manager/src/index.ts
```diff
@@ -0,0 +1,207 @@
...
+  ): Promise<string> {
+    const txHash = await this.transactionSender.sendTransaction(tx, validator);
+    await this.txStorage.addTransaction({ ...tx, hash: txHash });
+    return txHash;
+  }
+
+  async collector(
+    options: CKBIndexerQueryOptions,
+    usePendingCells = true
+  ): Promise<CellCollector> {
```
homura: keep the compatible here with the other `indexer.collector(): CellCollector`

```ts
  collector(
    queryOptions?: QueryOptions,
    options?: { usePendingOutputs?: boolean }
  ): CellCollector;
```

packages/transaction-manager/examples/chained-transfer-example/src/index.ts
zhangyouxin: Tried to use `common.transfer` but just returns empty tx, maybe wrong usage, I'll look into the source code of `common.transfer` and try again. Basiclly the function `transferToMySelf` in example is just implementing `common.transfer`

packages/transaction-manager/README.md
zhangyouxin: updated

packages/transaction-manager/src/index.ts
zhangyouxin: updated

packages/transaction-manager/examples/chained-transfer-example/src/index.ts
zhangyouxin: Another reason why fail to use `common.transfer` is that `TransactionManager.collector` reutrns `Promise<CellCollector>` which is not compatible to `CellCollector`, because `TransactionManager` needs to read from `storage`, which is async storage, so it returns a promise

packages/transaction-manager/README.md
zhangyouxin: updated

packages/transaction-manager/src/index.ts
homura: I find this is not updated, if it is updated, the [`common.transfer`](https://github.com/ckb-js/lumos/pull/503#discussion_r1188022100) will work. We can move the async code into `Collector.collect` instead of constructing a `Collector` in async, the `Indexer.collector` method is actually semantically constructing a `Collector`, and we should make sure the `construct(queryOptions)` be sync

zhangyouxin: ok，then I'll pass `txStorage` to `TransactionManagerCollector` class, and do async storage    access when `collect` is called

packages/transaction-manager/src/index.ts
zhangyouxin: updated

packages/transaction-manager/src/PendingTransactionStorage.ts
zhangyouxin: updated

packages/transaction-manager/src/index.ts
zhangyouxin: updated

packages/transaction-manager/src/index.ts
zhangyouxin: updated for code/test/readme/example

homura:
Vercel build failed, @zhangyouxin can you have a loot at it

```
Error: /vercel/path0/packages/transaction-manager/lib/TransactionStorage.d.ts(2)
--
14:37:20.291 | Cannot find name 'TransactionStorageType'.
14:37:20.291 | Error: /vercel/path0/packages/transaction-manager/lib/index.d.ts(2)
14:37:20.291 | Module '"./TransactionStorage"' has no exported member 'TransactionStorageType'.
14:37:20.346 | error Command failed with exit code 4.
14:37:20.347 | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
14:37:20.381 | Error: Command "npm run build-lumos && npm run build" exited with 4
14:37:20.681 | Deployment completed
14:37:20.637 | BUILD_UTILS_SPAWN_4: Command "npm run build-lumos && npm run build" exited with 4
```

homura (approved):
LGTM

homura (approved):
LGTM

PR #549, opened 2023-08-04
fix(common-scripts): sudt data may over 16 bytes

### LINKED ISSUE #548: Predefined LockScriptInfos cannot be replaced
**Describe the bug**

Not sure if this is a feature, please help me identify it.

Firstly, let's check the code:

https://github.com/ckb-js/lumos/blob/6bb442cb864484efa40020860a254cac219427e2/packages/common-scripts/src/common.ts#L80-L87

You can see the predefined LockScriptInfos and the custom LockScriptInfos are just two list added up together, without filters or anything. So what should I do if I don't want to use some of the predefined logic in `predefinedInfos`?

For instance, if I want to write my own `secp256k1Blake160` LockScriptInfo, how can I cover/replace the predefined one?

homura:
# Description

Fixes #548

the data of an sUDT cell may be over 16 bytes but it is still valid, so we should slice the first 16 bytes as the sUDT amount

## Type of change

### DIFF
--- a/.changeset/orange-flowers-sort.md
+++ b/.changeset/orange-flowers-sort.md
@@ -1,0 +1,5 @@
+---
+"@ckb-lumos/common-scripts": patch
+---
+
+enable ACP to work with cells with more than 16 bytes of data
--- a/.changeset/orange-flowers-sort.md
+++ b/.changeset/orange-flowers-sort.md
@@ -2,4 +2,4 @@
 "@ckb-lumos/common-scripts": patch
 ---
 
-enable ACP to work with cells with more than 16 bytes of data
+enable ACP to work with sUDT cells with more than 16 bytes of data
--- a/examples/exchange-sudt-for-ckb/package.json
+++ b/examples/exchange-sudt-for-ckb/package.json
@@ -15,10 +15,10 @@
     "process": "^0.11.10"
   },
   "dependencies": {
-    "@ckb-lumos/base": "next",
-    "@ckb-lumos/bi": "next",
-    "@ckb-lumos/codec": "next",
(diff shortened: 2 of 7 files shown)


ShookLyngs:
Not an issue, just out of curious: does the changesets need a strong format?

For example, I see the lumos' changeset messages are like this: "doing something".
What happens if it is like this: "Doing something"?
Or like this: "fix: doing something"?

I'm also using changesets, wonder if it matters.

homura:
> Not an issue, just out of curious: does the changesets need a strong format?
> 
> For example, I see the lumos' changeset messages are like this: "doing something". What happens if it is like this: "Doing something"? Or like this: "fix: doing something"?
> 
> I'm also using changesets, wonder if it matters.

No, the changesets has no restrictions on the format, here is an [example](https://github.com/ckb-js/lumos/pull/545)

.changeset/orange-flowers-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ckb-lumos/common-scripts": patch
+---
+
+enable ACP to work with cells with more than 16 bytes of data
```
zhangyouxin: enable ACP to work with sUDT cells with more than 16 bytes of data

zhangyouxin:
better add test case in `packages/common-scripts/tests/sudt.test.ts` though it's a small change

zhangyouxin:
Do we need to update the example: `examples/exchange-sudt-for-ckb/src/lib.ts`

.changeset/orange-flowers-sort.md
homura: updated

zhangyouxin (approved):

PR #552, opened 2023-08-15
 feat(config-manager): generate script config from genesis

### LINKED ISSUE #510: Is it possible to provide a function to generate the lumos config for dev net?

When I develop in dev net which I run the ckb node myself, I found that I need to find the script outpoint of `Secp256k1Blake160`, `Secp256k1Blake160Multisig` and `DAO` myself to setup the lumos config.

Maybe lumos could provider a function for developers like me to generate the config, such like `getSecp256k1Blake160Config()`.

**Describe alternatives you've considered**

**Additional context**

homura:
# Description

Resolve #510

## Example

```js
import { generateGenesisScriptConfigs } from '@ckb-luoms/config-manager'
import { RPC } from '@ckb-lumos/rpc'

const rpc = new RPC('http://127.0.0.1:8114')
const genesisBlock = rpc.getBlockByNumber('0x0')

const scriptConfigs = generateGenesisScriptConfigs(genesisBlock)
console.log(
  scriptConfigs.SECP256K1_BLAKE160,
  scriptConfigs.SECP256K1_BLAKE160_MULTISIG,
  scriptConfigs.DAO,
)
```

## Type of change


packages/config-manager/src/genesis.ts
```diff
@@ -0,0 +1,70 @@
+import { Block, utils } from "@ckb-lumos/base";
+import { ScriptConfig } from "./types";
+
+// https://github.com/nervosnetwork/ckb-sdk-rust/blob/94ce4379454cdaf046f64b346e18e73e029f0ae6/src/constants.rs#L19C1-L24C62
```
zhangyouxin, line 4: maybe we don't need to fully import type here like `usize` in rust.

```
type TransactionIndex = number;
type InputIndex = number;
type SystemScriptLocationInfo = [TransactionIndex, InputIndex]
export const SIGHASH_OUTPUT_LOC: SystemScriptLocationInfo = [0, 1];
```
seems more readable

homura: improved the readability

zhangyouxin (approved):

packages/config-manager/src/genesis.ts
```diff
@@ -0,0 +1,74 @@
...
+  ScriptConfig
+> {
+  if (!genesisBlock) throw new Error("cannot load genesis block");
+
+  return {
+    SECP256K1_BLAKE160: {
+      CODE_HASH: utils.computeScriptHash(
+        genesisBlock.transactions[SIGHASH_OUTPUT_LOC[0]].outputs[
+          SIGHASH_OUTPUT_LOC[1]
+        ].type!
```
zhangyouxin: The input parameter is `genesisBlock: Block`, which is assuming user to pass in a genesis block here. 

But we should handle a situation when user feeds a non-genesis block, here using `genesisBlock.transactions.outputs[x].type!` it will throw error like: `invalid script`.

It will be better to throw `Not a genesis block, please check the input block number` like, then add a test case to cover that

PR #560, opened 2023-10-11
refactor: remove peerDependency of toolkit

zhangyouxin:
# Description

This PR removes `peerDependency` of `toolkit` package

## Type of change

# How Has This Been Tested?

- [x] Smoke test to make sure base `Normalizers` using `normalizeHexNumber` works


homura:
We don't need to bump the version manually, can you follow the [For Contributor](https://github.com/ckb-js/lumos/blob/develop/docs/DEVELOPMENT.md#to-contributor) section to add the changeset?

packages/toolkit/src/normalizers.js
```diff
@@ -10,15 +10,15 @@
...
+import { BI } from "@ckb-lumos/bi";
 import { Reader } from "./reader";
 /**
  * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide}
  */
 function normalizeHexNumber(length) {
   return function (debugPath, value) {
     if (!(value instanceof ArrayBuffer)) {
-      let intValue = JSBI.BigInt(value).toString(16);
+      let intValue = BI.from(value).toHexString();
```
homura: they are not the same, because the `BI.toHexString()`'s result is `0xabcd` while `JSBI.toString(16)`'s result is `abcd`. Could you provide the reason for doing this?

zhangyouxin: fixed and added test case by [fe92d3e](https://github.com/ckb-js/lumos/pull/560/commits/fe92d3eb587483ab3ce319113f56511fbabc7b05), the motivation is that since we are implementing our own Bignumber lib in lumos, why not use is other than use third-party lib. Another reason is that peer dependency needs more configuration when used by other projects if they are using yarn as package manager https://github.com/nervosnetwork/neuron/pull/2843#issuecomment-1730907381 , if we remove peer dependency, the problem will be gone for other projects too

packages/toolkit/src/normalizers.js
```diff
@@ -10,15 +10,15 @@
...
+import { BI } from "@ckb-lumos/bi";
 import { Reader } from "./reader";
 /**
  * @deprecated please follow the {@link https://lumos-website.vercel.app/migrations/migrate-to-v0.19 migration-guide}
  */
 function normalizeHexNumber(length) {
   return function (debugPath, value) {
     if (!(value instanceof ArrayBuffer)) {
-      let intValue = JSBI.BigInt(value).toString(16);
+      let intValue = BI.from(value).toHexString().substring(2);
```
homura:
```suggestion
      let intValue = BI.from(value).toString(16);
```

homura (approved):

PR #564, opened 2023-11-08
refactor: migrate axios to fetch

### LINKED ISSUE #487: Use one HTTP client instead of mixed use of multiple clients
The `@ckb-lumos/ckb-indexer`, `@ckb-lumos/light-client` using the `cross-fetch` as an HTTP client, but the `@ckb-lumos/rpc` is using Axios since it is migrated from `ckb-sdk-js`.

We need to refactor `@ckb-lumos/rpc` from the axios to the fetch

## Why

We found that Lumos does not run very well in the Service Worker because the default adapter of Axios uses `XMLHttpRequest`, but Service Worker only supports fetch API as an HTTP client.

Lumos currently uses fetch by `import fetch from 'cross-fetch'` instead of `import 'cross-fetch'` to avoid mount `fetch` to `global`. But event if Lumos is refactored to `fetch`, it stil can't be made to run directly in the Service Worker, because of it still using fetch like `import {default as fetch} from 'cross-fetch'`, and the `cross-fetch` is based on the `XMLHttpReqeust`. However, most projects today are bundled via webpack, so we can add a plugin to make the Lumos run in the Service Worker

```ts
plugins: [
    new webpack.NormalModuleReplacementPlugin(
        /cross-fetch/, 
        require.resolve('./my-cross-fetch.js'),
    ),
]
```

```ts
// my-cross-fetch.js
module.exports = global.fetch
module.exports.default = global.fetch 
module.exports.fetch = global.fetch
module.exports.Headers = global.Headers
module.exports.Request = global.Request
module.exports.Response = global.Response
```

homura:
## Description

Fixes #487

This pull request updates the `@ckb-lumos/rpc` library to use `fetch` instead of `axios`. This change was made to avoid modifying `axios`'s adapter, which could potentially cause issues. By choosing a dependency without the state, Lumos can function better as a library. The `fetch` is not stateful, and it's widely supported in several runtimes, including the modern browsers, NodeJS, Deno, and Bun

Recently, a project using Axios and Lumos together experienced build failures, resulting in an error message that reads: 

```
ERROR in ./node_modules/@vespaiach/axios-fetch-adapter/index.js 2:0-43
Module not found: Error: Package path ./lib/core/settle is not exported from package 
```

This error was caused by https://github.com/ckb-js/lumos/blob/a243f118a1f5aff91ea282eebd3eff149fa562b1/packages/rpc/src/initAxiosWebworkerAdapter.ts#L15-L16

You may notice the `httpAgent` and `httpsAgent` were removed from the creation of axios in this pull request, but don't worry, they have never been used in Lumos

https://github.com/ckb-js/lumos/blob/a243f118a1f5aff91ea282eebd3eff149fa562b1/packages/rpc/src/index.ts#L23-L25

## Type of change

# How Has This Been Tested?

- [x] unit test


zhangyouxin (approved):
Looks like this PR also resolves axios adapter problem when making some http(s) API calls in chrome extension environment also?

homura:
> Looks like this PR also resolves axios adapter problem when making some http(s) API calls in chrome extension environment also?

you're right, it also works well in ServiceWorker

homura:
> You may notice the `httpAgent` and `httpsAgent` were removed from the creation of axios in this pull request, but don't worry, they have never been used in Lumos
> 
> https://github.com/ckb-js/lumos/blob/a243f118a1f5aff91ea282eebd3eff149fa562b1/packages/rpc/src/index.ts#L23-L25
> 

I suddenly realized it is wrong, the `http(s)Agent` may be passed through `setNode`

https://github.com/ckb-js/lumos/blob/68e4f5bf9e7ed4ba15c1ced0d3054c2a0be777a6/packages/rpc/src/index.ts#L66-L69

so that we still need to support the `http(s)Agent`. To support the agent, Lumos should make the `fetch` customizable, like this

```ts
// example for working with keepalive
const customizedFetch: typeof fetch = (request, init) => {
  return globalThis.fetch(request, {
    ...init,
    keepalive: true,
  });
};


const rpc = new RPC(url, { fetch: customizedFetch })
```


```ts
// or if you are still working with NodeJS
import fetch from 'node-fetch'

new PRC(url, {
  fetch: (request, init) => {
    return fetch(request, { ...init, agent: customAgent })
  }
})

```

PR #588, opened 2023-12-16
refactor!: make `@ckb-lumos/lumos` more convenient

### LINKED ISSUE #585: Too many packages need to upgrade after lumos upgrading
Lumos has more than 15 scoped child packages `@ckb-lumos/*` in it, although Lumos provides an entry package `@ckb-lumos/lumos` to integrate them into one, however, it still doesn't meet its target of just importing one entry package, because not all exported modules in child packages are included.

We need to improve the `@ckb-lumos/lumos` to make the entry package more convenient for developers, so they don't need to import too many `@ckb-lumos/*`

homura:
# Description

Resolves #585 

Go to the [migration](https://github.com/ckb-js/lumos/blob/25c4ed27efbc17512520f2324d3751ab3e726d60/website/docs/migrations/migrate-to-v0.22.md) for more details

## Type of change

- [x] Example

# How Has This Been Tested?

- [x] Use [Verdaccio](https://github.com/homura/lumos-experimental/actions/runs/7245400144/job/19735357478) to publish the test pkgs


examples/exchange-sudt-for-ckb/src/lib.ts
```diff
@@ -1,12 +1,9 @@
-import { Indexer, helpers, Address, Script, RPC, hd, config, Cell, commons, BI } from "@ckb-lumos/lumos";
-import { sudt } from "@ckb-lumos/common-scripts";
-import { BIish } from "@ckb-lumos/bi";
-import { payFeeByFeeRate } from "@ckb-lumos/common-scripts/lib/common";
-import { addCellDep } from "@ckb-lumos/common-scripts/lib/helper";
+import { Indexer, helpers, Address, Script, RPC, hd, config, Cell, commons, BI, BIish } from "@ckb-lumos/lumos";
+import { sudt, common } from "@ckb-lumos/lumos/common-scripts";
+import { addCellDep } from "@ckb-lumos/lumos/helpers";
+import { hexify, Uint128, blockchain } from "@ckb-lumos/lumos/codec";
+import { computeScriptHash } from "@ckb-lumos/lumos/utils";
```
Keith-CY, line 5: Can these be imported via `@ckb-lumos/lumos`?

homura: I am not sure if it is good to export many functions directly in `@ckb-lumos/lumos`. 

These lines can be refactored to

```ts
import { helpers, utils, commons } from '@ckb-lumos/lumos'
import { Uint128, hexify, blockchain } from '@ckb-lumos/lumos/codec'

const { addCellDep } = helpers
const { computeScriptHash } = utils
const { sudt, common } = commons
```

---

If all child pieces, such as functions, classes, or event interfaces are exported in the entry package root, the list of hints becomes very long. So only some common functions are exported now. Another reason for doing this is to make the modules keep the same style, to avoid

```ts
import { addCellDep, helper } from '@ckb-lumos/lumos'

helper.addCellDep(...)
// ...
addCellDep(...)
````

But I think it's OK for doing something like this

```
import { addCellDep } from '@ckb-lumos/lumos/helper'

addCellDep(...)
```

Because for someone who is not the direct author of the code, it is easy to realize that the `addCeppDep` is a helper-related method, i.e. a function used to assemble a transaction(although `helper` is not a very good package name for assembling transactions, but we still use the name to keep the compatibility)

**Comment by homura** [comment]
@zhangyouxin can you help to check this PR?

File: packages/lumos/clean.mjs [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+  ".npmignore",
+  "CHANGELOG.md",
+  "clean.mjs",
+  "package.json",
+  "README.md",
+  "tsconfig.json",
+  "typedoc.json",
+];
+
+function removeFilesExceptWhitelist(dir) {
```
zhangyouxin, line 20: What directories other than `dist` or `build` are expected to be cleaned up by this script? is it convenient to just add a script in `package.json`, like `clean: rimraf ./build`

homura: The new build script now outputs all built files in the `packages/lumos` directory, making it more convenient for developers to import modules without a `/lib` or `/dist` when the `moduleRelution` is set to the default `node10` or `node`. Developers can import directly like this:

```js
import { secp256k1Blake160 } from '@ckb-lumos/lumos/common-scripts'
import { Uint8, Uint16, Uint32 } from '@ckb-lumos/lumos/codec'

// So developers needn't import modules like the following way
import { secp256k1Blake160 } from '@ckb-lumos/lumos/lib/commons-scripts'
```

This is more natural, and it's smoother to upgrade to the ESM in the future

zhangyouxin (approved):

examples/exchange-sudt-for-ckb/src/lib.ts
Keith-CY: Make sense

Keith-CY (approved):

PR #599, opened 2024-01-05
refactor: allow Transaction.pack(unpacked)

doitian:
# Description

The function can pack its unpacked result, but the function signature
does not allow it. This is very annoying, for example, I cannot set
version to number 0 to call `pack`.

## Type of change

# How Has This Been Tested?

- [x] added a test case in `packages/base/tests/serialize.test.ts`
- [x] `pnpm --filter base lint && pnpm --filter base test`


homura:
Thanks for your PR. However, I have noticed that the `blockchain.Transaction.unpack` function behaves differently from `api.Transaction`. The `blockchain` module directly uses number codecs, which results in the unpacked output being either a `number` or a `BI`. Ideally, any type of number should be in `string(hex formatted)`. This inconsistency was not caused by your PR, but by your comment "cannot set number 0 to call pack", which drew my attention to it. To address this issue, Lumos may have to modify the `blockchain.*.unpack` function to output a `string(hex format)` instead of a `number`.

homura:
@doitian Hi, I've opened a new PR that should solve this issue at #601, please let me know if you think it helps

doitian:
> @doitian Hi, I've opened a new PR that should solve this issue at #601, please let me know if you think it helps

Not sure whether it worthes a breaking change.

homura:
> Not sure whether it worthes a breaking change.

This breaking change is not just a fix for this PR, it is also a bug since 0.20

PR #601, opened 2024-01-06
BREAKING CHANGE: unexpected unpacked result in blockchain

### LINKED ISSUE #600: Some of the results of blockchain.*.unpack is inconsistent with `api.*`
Thanks for your PR. However, I have noticed that the `blockchain.Transaction.unpack` function behaves differently from `api.Transaction`. The `blockchain` module directly uses number codecs, which results in the unpacked output being either a `number` or a `BI`. Ideally, any type of number should be in `string(hex formatted)`. This inconsistency was not caused by your PR, but by your comment "cannot set number 0 to call pack", which drew my attention to it. To address this issue, Lumos may have to modify the `blockchain.*.unpack` function to output a `string(hex format)` instead of a `number`.

_Originally posted by @homura in https://github.com/ckb-js/lumos/issues/599#issuecomment-1878777415_

homura:
# Description

Fixes #600

This pull request fixed the unpacked result of `blockchain.*.unpack`. It now matches the RPC type, which uses hex format instead of `number` or `BI`.

```ts
const { version } = blockchain.Transaction.unpack('0x...')

// before 
typeof version === 'number' // ❌ 
version === 0 // ❌ 

// after
typeof version === 'string' // ✅
version === '0x0' // ✅
```

This bug was caused by lumos@0.20 at #413 but was never reported before #599 because the API `Transaction.unpack` or `Block.unpack` was so rarely used.

## Type of change

# How Has This Been Tested?

- [x] Unit tests involve both `pack()` and `unpack()` in `base.blockchain`

### DIFF
--- a/.changeset/rare-mugs-begin.md
+++ b/.changeset/rare-mugs-begin.md
@@ -2,4 +2,4 @@
 "@ckb-lumos/base": minor
 ---
 
-BREAKING CHANGE: unpacked number of blockchain.\*.unpack should be `Hexadecimal` instead of `number` or`BI`
+BREAKING CHANGE: unpacked number of `blockchain.*.unpack` should be `Hexadecimal` instead of `number` or`BI`
--- a/.changeset/rare-mugs-begin.md
+++ b/.changeset/rare-mugs-begin.md
@@ -1,0 +1,5 @@
+---
+"@ckb-lumos/base": minor
+---
+
+BREAKING CHANGE: unpacked number of blockchain.\*.unpack should be `Hexadecimal` instead of `number` or`BI`
--- a/packages/base/src/blockchain.ts
+++ b/packages/base/src/blockchain.ts
@@ -12,9 +12,22 @@ import {
 import { BytesCodec, FixedBytesCodec } from "@ckb-lumos/codec/lib/base";
 
 import type * as api from "./api";
-import { BI } from "@ckb-lumos/bi";
+import { BI, BIish } from "@ckb-lumos/bi";
 
-const { Uint128LE, Uint8, Uint32LE, Uint64LE } = number;
+function asHexadecimal(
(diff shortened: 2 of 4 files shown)


Keith-CY:
It works but the solution seems weird, it might be a historical issue but I still wonder why is `number` allowed when the risk of `number` is well-known

homura:
> It works but the solution seems weird, it might be a historical issue but I still wonder why is `number` allowed when the risk of `number` is well-known

I found that the problem originated from a large PR at #413. The PR removed the hex format [conversion](https://github.com/ckb-js/lumos/pull/413/files#diff-6e024e513dee3184b12b2da0fc366e84a3860b23b38e13726643beaf809894b8L355), but the reviewer didn't realize that the return value type had changed. This was because the types were annotated by being inferred from `type Transaction = UnpackResult<typeof SomeCodec>`. And there are rare scenarios in which `Transaction.unpack` is used during development because developers typically use the the `Transaction` with the [default `verbosity`](https://github.com/nervosnetwork/ckb/blob/develop/rpc/README.md#method-get_transaction) instead of the serialized one

Keith-CY (approved):

Keith-CY:
Approved but I'm still worried about the usage of `number`. My opinion might be kinda extreme, the `number` should be replaced by `string` and `big number`

homura:
> Approved but I'm still worried about the usage of `number`. My opinion might be kinda extreme, the `number` should be replaced by `string` and `big number`

It might be not consistent with some blockchain SDK, such as [typechain](https://github.com/dethcrypto/TypeChain/blob/4c9773403a2aa66bf6f5fa066e8ef1b00769c71b/packages/target-ethers-v5/src/codegen/types.ts#L61-L62). Lumos's current default number codec draws on some developer conventions from them

PR #602, opened 2024-01-08
fix(rpc): get_live_cell should be nullable

homura:
# Description

[`get_live_cell` should be nullable](https://github.com/nervosnetwork/ckb/blob/0c33675e11ec072fbbe41c135a88749167f204d0/util/jsonrpc-types/src/cell.rs#L49) to force checking the `cell` field at compile time when using TypeScript

```ts
TypeError: Cannot read properties of null (reading 'output')
     28|   const liveCell = await props.rpc.getLiveCell(props.outPoint, true);
     29|   const cell: Cell = {
     30|     cellOutput: liveCell.cell.output,
       |                               ^
     31|     data: liveCell.cell.data.content,
     32|     outPoint: props.outPoint,
```

## Type of change

### DIFF
--- a/packages/rpc/src/Base/index.ts
+++ b/packages/rpc/src/Base/index.ts
@@ -130,7 +130,7 @@ export interface Base {
     outPoint: CKBComponents.OutPoint,
     withData: boolean
   ) => Promise<{
-    cell: CKBComponents.LiveCell;
+    cell: Option<CKBComponents.LiveCell>;
     status: CKBComponents.CellStatus;
   }>;
 
@@ -543,3 +543,5 @@ export class Base {
     return this.#rpcProperties;
   }
 }
+
+export type Option<T> = T | null;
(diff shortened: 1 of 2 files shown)


packages/rpc/src/Base/index.ts
```diff
@@ -543,3 +543,5 @@ export class Base {
     return this.#rpcProperties;
   }
 }
+
+export type Option<T> = T | null;
```
zhangyouxin, line 547: `Nullable` is more semantic

homura: `Option<T>` here is for keeping the same with the `jsonrpc-types` https://github.com/nervosnetwork/ckb/blob/0c33675e11ec072fbbe41c135a88749167f204d0/util/jsonrpc-types/src/cell.rs#L49C15-L49C21

zhangyouxin (approved):

PR #606, opened 2024-01-11
feat: deprecate config/helpers and fix website address conversion

homura:
# Description

This PR

1. deprecated the submodule `helpers` in `@ckb-lumos/lumos/config` and `@ckb-lumos/config-manager` to avoid confusion with `@ckb-lumos/helpers`
2. fixed the address conversion 

```ts
import { helpers } from '@ckb-lumos/lumos/config' // Not Recommended
import { helpers } from '@ckb-lumos/config-manager' // Not Recommended
import { nameOfScript } from '@ckb-lumos/lumos/config' // Recommended
```

## Type of change

### DIFF
--- a/packages/lumos/src/config.ts
+++ b/packages/lumos/src/config.ts
@@ -16,4 +16,13 @@ export {
   validateConfig,
   initializeConfig,
   getConfig,
+  /**
+   * @deprecated use the {@link nameOfScript} and {@link findConfigByScript} function instead
+   */
+  helpers,
 } from "@ckb-lumos/config-manager";
+
+export {
+  nameOfScript,
+  findConfigByScript,
+} from "@ckb-lumos/config-manager/lib/helpers";
--- a/.changeset/wild-dodos-notice.md
+++ b/.changeset/wild-dodos-notice.md
@@ -1,0 +1,6 @@
+---
+"@ckb-lumos/config-manager": minor
+"@ckb-lumos/lumos": minor
+---
+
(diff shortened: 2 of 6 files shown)


packages/lumos/src/config.ts
```diff
@@ -16,4 +16,13 @@ export {
   validateConfig,
   initializeConfig,
   getConfig,
+  /**
+   * @deprecated use the {@link nameOfScript} and {@link findConfigByScript} function instead
+   */
+  helpers,
```
Keith-CY, line 22: Why not add the mark where `helpers` is defined

homura: The submodule named `helpers` is defined in `@ckb-lumos/config-manager/lib/helpers`, and there was a `helpers` already in `@ckb-lumos/helpers`, this may confuse.

```ts
import { helpers } from '@ckb-lumos/lumos' // OK

import { helpers } from '@ckb-lumos/lumos/config' // Not recommended
import { nameOfScript } from '@ckb-lumos/lumos/config' // Recommended

```

Keith-CY: I mean why not mark `deprecated` above `helpers` in `@ckb-lumos/lumos/config`

homura: you're right, the original helpers should also be marked as deprecated for more clear

Keith-CY (approved):

packages/lumos/src/config.ts
homura: @Keith-CY Can you mark this comment as resolved if you think it's OK?

PR #607, opened 2024-01-11
feat: supported bitcoin auth in omnilock

### LINKED ISSUE #604: Support new auth flag bitcoin via unisat in Omnilock
The latest Omnilock supported a new auth flag called to enable verifying signature signed from UniSat or OKXWallet.

- [UniSat `signMessage`](https://docs.unisat.io/dev/unisat-developer-service/unisat-wallet#signmessage)
- [Omnilock Integration PR](https://github.com/nervosnetwork/ckb-production-scripts/pull/75)

homura:
# Description

Fixes #604 

This PR supported Bitcoin auth in Omnilock

```ts
import { bitcoin, createOmnilock } from '@ckb-lumos/lumos/common-scripts'

// create the corresponding Omnilock-Bitcoin lock
const lock = createOmnilock({
  auth: { flag:'BITCOIN', address: (await window.unisat.requestAccounts)[0] }
})

// sign the transaction
const message = txSkeleton.get('signingEntries').get(n).message
const signature = bitcoin.signMessage(message, 'ecdsa', window.unisat)
const sealedTx = sealTransaction(txSkeleton, [signature])
```

## Example

```
git checkout origin/omnilock-bitcoin-auth
pnpm install
pnpm build
cd examples/omni-lock-unisat
pnpm install
npm start
```

ckb-via-unisat

## Type of change

# How Has This Been Tested?

- [x] unit test
- [x] example


packages/common-scripts/src/omnilock-bitcoin.ts
```diff
@@ -0,0 +1,93 @@
+import { bytes } from "@ckb-lumos/codec";
+import { bech32 } from "bech32";
+import bs58 from "bs58";
+
+export function decodeAddress(address: string): ArrayLike<number> {
+  try {
+    // Bech32
+    if (address.startsWith("bc1")) {
```
zhangyouxin: both Native Segwit and Taproot addresses start with `bc1` here

homura: the taproot address is encoded via bech32m so it will fail when trying to decode via bech32, the following catch block will throw an error to tell the user that the p2tr is not supported yet

homura: the taproot address is encoded via bech32m so it will fail when trying to decode via bech32, the following catch block will throw an error to tell the user that the p2tr is not supported yet

homura:
Hold on, a new concept ["display message"](https://github.com/nervosnetwork/ckb-production-scripts/pull/75/commits/3bc30c8d370ebc824d46b29de8151e1a2a6ea638) comes to Omnilock that prepends a human-readable prefix before the transaction digest, like `CKB (Bitcoin Layer-2) transaction: <digest>`. So we should update the signing part in this PR

examples/omni-lock-unisat/README.md
```diff
@@ -0,0 +1,21 @@
+# Lumos Works with UniSat
+
+Nervos maintains a powerful lock
+called [Omni Lock](https://github.com/XuJiandong/docs-bank/blob/master/omni_lock.md) (previously named RC lock), which
```
Keith-CY: Why not link it to the RFC at https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0042-omnilock/0042-omnilock.md

examples/omni-lock-unisat/README.md
```diff
@@ -0,0 +1,21 @@
...
+can use MetaMask as a signer. This example will show how to use Lumos to send a transaction using Omni lock and MetaMask
```
Keith-CY: The spelling of `Omni Lock` should be unified, it's called `Omni Lock` at line 4 but `Omni lock` at this line.

BTW, it's called `Omnilock` in the RFC at https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0042-omnilock/0042-omnilock.md

packages/common-scripts/src/omnilock.ts
```diff
@@ -66,17 +88,21 @@ export function createOmnilockScript(
...
-    // 01: Ethereum     👇            01: administrator
-    if (omnilockInfo.auth.flag === "ETHEREUM") {
-      return `0x01${bytes.hexify(omnilockInfo.auth.content).slice(2)}00`;
-    }
-    if (omnilockInfo.auth.flag === "SECP256K1_BLAKE160") {
-      return `0x00${bytes.hexify(omnilockInfo.auth.content).slice(2)}00`;
+    const flag = omnilockInfo.auth.flag;
+    switch (flag) {
+      case "ETHEREUM":
+        return bytes.hexify(bytes.concat([1], omnilockInfo.auth.content, [0]));
```
Keith-CY: `0`, `1`, `4` become magic numbers without the comment.

packages/common-scripts/src/omnilock.ts
```diff
@@ -35,17 +35,39 @@ import {
...
+};
+export type IdentityBitcoin = {
+  flag: "BITCOIN";
+  /**
+   * a Bitcoin address, such as
+   * `P2PKH(17VZNX1SN5NtKa8UQFxwQbFeFc3iqRYhem)`,
+   * `P2SH(3EktnHQD7RiAE6uzMj2ZifT9YgRrkSgzQX)`,
+   * `Bech32(bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4)`
+   */
+  address: string;
```
Keith-CY: IdentityEthereum at https://github.com/ckb-js/lumos/pull/607/files#diff-64dca3ead622f318b87b2ea50ea9486f3aaad456a9d4325106f0871c2a4af1cfR60 has a field `content` with an ethereum address as its value

Why use a different field `address` for the IdentityBitcoin

packages/common-scripts/src/omnilock-bitcoin.ts
```diff
@@ -0,0 +1,97 @@
...
+  const address = accounts[0];
+  if (address.startsWith("bc1q")) {
+    signature[0] = 39 + ((signature[0] - 27) % 4);
+  }
+  if (address.startsWith("3")) {
+    signature[0] = 35 + ((signature[0] - 27) % 4);
+  }
+  if (address.startsWith("1")) {
+    signature[0] = 31 + ((signature[0] - 27) % 4);
+  }
```
Keith-CY: `39`, `35`, `31` are all magic numbers

packages/common-scripts/src/omnilock.ts
homura: This is because the Bitcoin browser wallets, such as UniSat and OKX Wallet, return a BitCoin address instead of a public key hash directly. To ensure the `createOmnilock` is easy to use, we've converted addresses into public keys silently via `bitcoin.decodeAddress`.

Keith-CY: What confused me is that thare are 2 fields in `IdentityEthereum` and `IdentityBitcoin`
- flag
- content | address
- 
As mentioned at line 58, `content` in `IdentityEthereum` is also an address, so why not use the same field, `address`, or both use `content`

homura: > why not use the same field, address, or both use content

the content here means the public key hash. However, the public key hash is an alias of the address on Ethereum.

Maybe we can change it to 

```ts
type IdentityEthereum = { flag: "Ethereum" } & ({content: BytesLike} | {address: BytesLike})
```

This would be more clear

Keith-CY: I have no idea how to make it clearer, maybe we can leave it for now.

homura: I've unified to use `content` to make it easier to use

packages/common-scripts/src/omnilock-bitcoin.ts
homura: comment added

packages/common-scripts/src/omnilock.ts
homura: commented in the latest commit

Keith-CY (approved):

PR #610, opened 2024-01-12
feat(config-manager): supported refresh deployment config

### LINKED ISSUE #609: Providing a method to automatically detect and upgrade outdated contracts configuration
The testnet Omnilock has recently got an upgrade to support more auth modes. This has resulted in certain dApps on the testnet experiencing unexpected issues, as the new deployment location `TX_HASH` has replaced the old one when sending transactions using Omnilock. Therefore, it would be better if Lumos provided a method to detect and update the corresponding configuration to avoid failure in the production environment.

The method could be something like this

```ts
interface Fetcher {
  (txHash: string) => PromiseLike<Transaction>;
}

declare function upgrade<S extends ScriptConfigs>(scriptConfigs: S, fetcher: Fetcher): Promise<S>;
```

homura:
# Description

Fixes #609

This PR provided a method `refreshScriptConfigs` to refresh deployment configs for `ScriptConfigs`, developers can use the method to refresh configs, such as the `OMNILOCK` in `predefined.AGGRON4.SCRIPTS`

```ts
const refreshed = await refreshScriptConfigs(
  predefined.AGGRON4.SCRIPTS,
  { resolve: createRpcResolver(rpc) }
);
initializeConfig({SCRIPTS: refreshed, PREFIX: 'ckt'});
```

## Type of change

# How Has This Been Tested?

- [x] unit test

### DIFF
--- a/packages/config-manager/src/refresh.ts
+++ b/packages/config-manager/src/refresh.ts
@@ -112,7 +112,7 @@ type RefreshConfig<S> = {
 /**
  * Refreshing the config items in {@link ScriptConfigs} which are deployed with type id
  * @example
- * const updatedScriptConfigs = refreshScriptConfigs(predefined.AGGRON4.SCRIPTS, createRpcResolver(rpc))
+ * const updatedScriptConfigs = await refreshScriptConfigs(predefined.AGGRON4.SCRIPTS, { resolve: createRpcResolver(rpc) })
  * initializeConfig({ SCRIPTS: updatedScriptConfigs })
  */
 export async function refreshScriptConfigs<S extends ScriptConfigs>(
--- a/packages/config-manager/src/refresh.ts
+++ b/packages/config-manager/src/refresh.ts
@@ -15,7 +15,7 @@ export type FetchOutputsByTxHashes = (txHashes: string[]) => MaybePromise<{ outp
 /**
  * fetch cells with corresponding type script
  */
-export type FetchOutPointByTypeId = (scripts: Script[]) => MaybePromise<{ outPoint: OutPoint }[]>;
+export type FetchOutPointsByTypeIds = (scripts: Script[]) => MaybePromise<{ outPoint: OutPoint }[]>;
 
 // prettier-ignore
 /**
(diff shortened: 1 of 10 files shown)


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
Keith-CY: Why is the version downgraded

packages/config-manager/src/refresh.ts
```diff
@@ -0,0 +1,175 @@
...
+ * resolve the latest `OutPoint[]` that has consumed the old `OutPoint[]`
+ */
+export type ResolveLatestOutPointsOfTypeIds = (outPoints: OutPoint[]) => MaybePromise<OutPoint[]>;
+// prettier-ignore
+export type FetchOutputsByTxHashes = (txHashes: string[]) => MaybePromise<{ outputs: Output[] }[]>;
+// prettier-ignore
+/**
+ * fetch cells with corresponding type script
+ */
+export type FetchOutPointByTypeId = (scripts: Script[]) => MaybePromise<{ outPoint: OutPoint }[]>;
```
Keith-CY: It's named with singular but a plural is returned

packages/config-manager/src/refresh.ts
```diff
@@ -0,0 +1,175 @@
...
+    });
+  };
+
+  const fetchIndexerCells: FetchOutPointByTypeId = async (typeIds) => {
+    const res = await rpc
+      .createBatchRequest<unknown[], CKBComponents.GetLiveCellsResult<false>>(
+        typeIds.map((typeId) => [
+          "getCells",
+          {
+            script: typeId,
```
Keith-CY: Maybe the argument `typeId` could be the args in the script because other fields are determined. And a string is more like an `Id`, IMO.

pnpm-lock.yaml
homura: Maybe there's something wrong with my local pnpm, I've rollback it,

Keith-CY (approved):

PR #621, opened 2024-01-21
feat: export parse from debugger

XuJiandong:
Changes:
The parseDebuggerData is useful to print json to stdout.  Fix a minor bug when ckb-debugger is missing.

Maybe this PR is not perfect. The requirement is to export `parseDebuggerData` or similar function. Feel free to use you own updates.

### DIFF
--- a/packages/debugger/src/executor.ts
+++ b/packages/debugger/src/executor.ts
@@ -86,6 +86,7 @@ export class CKBDebugger implements Executor {
         buf.stderr.toString("utf-8")
       );
     } else {
+      /* c8 ignore next 3 */
       throw new Error(
         `Failed to install ckb-debugger or ckb-debugger can't be located at ${this.debuggerPath}`
       );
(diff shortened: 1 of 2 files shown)


homura (approved):

packages/debugger/src/executor.ts
```diff
@@ -79,9 +79,16 @@ export class CKBDebugger implements Executor {
...
-      buf.stderr.toString("utf-8")
-    );
+    // when ckb-debugger can't be located, they're null.
+    if (buf.stdout != null && buf.stderr != null) {
+      return parseDebuggerMessage(
+        buf.stdout.toString("utf-8"),
+        buf.stderr.toString("utf-8")
+      );
+    } else {
+      throw new Error(
```
homura, line 90:
```suggestion
      /* c8 ignore next 3 */
      throw new Error(
```

PR #623, opened 2024-01-22
feat: update omnilock bitcoin sign message

homura:
# Description

Update the Omnilock Bitcoin signing message and the testnet config 

https://github.com/cryptape/omnilock/pull/2/commits/9419b7795641da0ade25a04127e25d8a0b709077

## Type of change

# How Has This Been Tested?

- [x] unit test
- [x] manual test

image
image


.changeset/honest-trainers-remain.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ckb-lumos/common-scripts": minor
+"@ckb-lumos/config-manager": minor
+---
+
+feat: omnilock bitcoin auth message is changed to `CKB (Bitcoin Layer) transaction: 0x`
```
Keith-CY, line 6: Is this a breaking change?

homura: the behavior here is more like a fixture since it doesn't work now if we don't update the magic PREFIX to be the same with the latest testnet Omnilock

Keith-CY: Got it

Keith-CY (approved):

PR #635, opened 2024-02-22
feat: codegen for molecule

### LINKED ISSUE #630: Public the child package `@ckb-lumos/parser` to work with custom molecule structures without manual binding
We've received feedback from the community that Lumos now is not convenient to work with custom molecule files since we need to write many pieces of binding.

So we need to public the child package `@ckb-lumos/molecule` and create some toolkit to make the  package working better with TypeScript

homura:
# Description

Fixes #630

This PR introduces a cli `lumos-molecule-codegen` to generate code from molecule schemas. 

Currently, there are some limitations of this codegen, such as

- the [`Molecule import statement`](https://github.com/nervosnetwork/molecule/blob/37748b1124181a3260a0668693c43c8d38c98723/docs/grammar/grammar.ebnf#L68-L70) is not supported yet

## Type of change

# How Has This Been Tested?

- [x] unit test
- [x] manually test via example


homura:
The first version of the codegen doesn't support the [import statement](https://github.com/nervosnetwork/molecule/blob/37748b1124181a3260a0668693c43c8d38c98723/docs/grammar/grammar.ebnf#L68-L70) of Molecule yet. To deal with the cases of importing other schemas, the codegen CLI also needs to 

1. resolve all import schema items
2. remove the intersection with `prepend`
3. append them to the `prepend`

PainterPuppets (approved):
Amazing work!

It seems that cli can only read `schemaFile` from config file, it would be better if it supports setting by `args`.

homura:
> It seems that cli can only read `schemaFile` from config file, it would be better if it supports setting by `args`.

Exactly, we can make the CLI easier to use later

PR #641, opened 2024-02-28
fix(debugger): conflict while running debugger in parallel

### LINKED ISSUE #625: conflict in debugger when executing transactions in parallel
**Describe the bug**

We should implement the following TODO to avoid the conflict when executing multiple transactions in parallel 

https://github.com/ckb-js/lumos/blob/00d74cd6bd06e876f3a45ff25fe44574c3bb3e61/packages/debugger/src/executor.ts#L54-L55

CI failed at the [attemp1](https://github.com/ckb-js/lumos/actions/runs/7607583096/attempts/1) and the [attemp2](https://github.com/ckb-js/lumos/actions/runs/7607583096/attempts/2) since omnilock bitcoin auth test in parallel mode

**Expected behavior**

CI pass without retry

## Notes

It is recommended to use a predictable name for the temp transaction file to ensure easy removal after the test is completed.

Web3TomCat:
Fixes #625 

This PR fixes is that when the user executes the debugger, a predictable temporary file is generated to avoid filename conflicts.

### DIFF
--- a/packages/debugger/src/executor.ts
+++ b/packages/debugger/src/executor.ts
@@ -6,6 +6,7 @@ import * as fs from "fs";
 import * as os from "os";
 import * as path from "path";
 import { parseDebuggerData, parseDebuggerMessage } from "./parse";
+import * as crypto from 'crypto';
 
 interface DebuggerOptions {
   readonly loader: DataLoader;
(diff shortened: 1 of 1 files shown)


packages/debugger/src/executor.ts
```diff
@@ -50,9 +51,9 @@ export class CKBDebugger implements Executor {
    */
   private saveTmpTxFile(txSkeleton: TransactionSkeletonType): string {
     const debuggerData = parseDebuggerData(txSkeleton, this.loader);
-
-    // TODO replace with random tmp file name to avoid conflict
-    const tmpTxPath = path.join(os.tmpdir(), "ckb_debugger_tx.json");
+    const randomHash = crypto.randomBytes(18).toString('hex');    
+    const tempFileName = `lumos-debugger-data-${randomHash}`
```
homura:
```suggestion
    const randomHex = crypto.randomBytes(18).toString('hex');    
    const tempFileName = `lumos-debugger-data-${randomHex}`    
```

"randomHex" would be a more suitable term since it does not refer to a hash.

Web3TomCat (commented):
cool

homura (approved):
LGTM

Issue #597, opened 2024-01-04
Make Lumos Work Smother With the Modern Frontend Dev Tools


homura:
Currently, Lumos has some methods, particularly crypto-related APIs such as `mnemonicToSeedSync(): Buffer` in `@ckb-lumos/hd`. This makes it a bit challenging to integrate Lumos directly with some modern developer tools like Vite. To work with Vite, developers typically require additional configuration to include these methods in the project, which is inconvenient.

I would like to propose a refactor to replace these methods with a modern one such as https://github.com/paulmillr/noble-curves, which is suitable for working with multiple major runtimes. Noble-curves is a trusted option for many projects as it has been audited. However, noble-curves relies on the `bigint` API, which means a breaking change for Lumos to use it as the NodeJS's `node:cyrpto` alternative.

Around two years ago, Lumos introduced a big integer library `@ckb-lumos/bi` and migrated the bigint-related API to the library for compatibility in #212, as some browsers didn't support `bigint`, especially Safari on iOS < 13.7.

Therefore, if we want to achieve this, some breaking changes will be required:

- Introduce `bigint` and drop the compatibility of iOS < 13.7([the market share < 3%](https://gs.statcounter.com/ios-version-market-share/) at 2023-12)
- Change the `Buffer` returned from some methods to `Uint8Array`

But this makes Lumos

- Smaller because we don't need the whole `browserify-crypto`, `browserify-buffer`, etc.
- More convenient because we don't need additional polyfill config in the project

Let me know if you think this change is worth it!

## Q&A

- Does the fallback of `Buffer` to `Uint8Array` affect massively?

Probably not, the `Buffer` is extended from `Uint8Array`, therefore, I suppose it is a slight change, or even without any change if developers do not use the `Buffer` API.

```diff
- const seed = mnemonicToSeedSync()
+ const seed = Buffer.from(mnemonicToSeedSync())
console.log(seed.toString('hex'))
```

- For the long-term perspective, how would this change impact BI?

It depends on developers' habits. I'd also like to add that the current BI supports [construct via bigint](https://github.com/ckb-js/lumos/blob/ccd821307e2bd9ddea7ab934c6b7cfb85762b3ee/packages/bi/src/index.ts#L2)

## Notes

- https://github.com/ckb-js/lumos/issues/570

jordanmack:
I support updating. The change sounds reasonable and the updates needed to support it are minimal.

[...]

PR #645, opened 2024-03-05
BREAKING CHANGE: replace `Buffer` with `Uint8Array`

homura:
# Description

This PR is a part implementation of https://github.com/ckb-js/lumos/issues/597. There are some breaking changes in this PR because all `Buffer` in parameters and returns of methods are replaced with `Uint8Array` for browser compatibility.

This PR targets the `browser-compatibility` branch instead of the `develop` branch because the feature may be released at 0.23.0

## Type of change

# How Has This Been Tested?

- [ ] Unit test


packages/hd/src/helper.ts
```diff
@@ -25,3 +26,11 @@ export function assertChainCode(chainCode: HexString): void {
     throw new Error(`chainCode must be length of 32 bytes!`);
   }
 }
+
+export function hexifyWithout0x(value: BytesLike): string {
```
Keith-CY, line 30: Why not extend the `bytes.hexify` to accept hex string without `0x`

twhy:
@Keith-CY @homura  After checking the related code last night, I suggest we postpone this PR.

Reasons
1. The `browser-compatibility` branch is 27 commits behind `develop`
2. This PR still contains `Buffer.from()` , e.g.
     https://github.com/ckb-js/lumos/pull/645/files#diff-a9e7dac36ed4d2aa4efe20eec63d01ded13115251c933f0dcfce2284d6e7024bR238
3. Replace `Buffer` with `Uint8Array` should be put behind replacing node crypto becuase it's a breaking change.

Our goal here is to make Lumos run in browsers without extra configs.
I want to achieve this with a few PRs based on `develop` branch.
1. [NON BREAKING] Replace Node `crypto.randomBytes` with Lumos `crypto.randomBytes`  [Code](https://github.com/ckb-js/lumos/blob/replace-node-crypto-randomBytes/packages/crypto/src/index.ts)  [PR](https://github.com/ckb-js/lumos/pull/701)
2. [NON BREAKING] Replace Node crypto `createHash` `createHmac` etc with [noble-curves](https://github.com/paulmillr/noble-curves) methods
3. [BREAKING] Replace `Buffer` with `Uint8Array` **fully**, based on existing work in this PR.
4. Maybe one more PR to make things fully work.

When all these done, we release it with version `0.30.0`
That means, starting from `0.30.0`, Lumos can run in browsers without extra configs.
People still want to use `Buffer` can stay at version `0.2x.x`

PR #646, opened 2024-03-05
Codegen 643

### LINKED ISSUE #643: Support the import statement of molecule for the codegen
> The first version of the codegen doesn't support the [import statement](https://github.com/nervosnetwork/molecule/blob/37748b1124181a3260a0668693c43c8d38c98723/docs/grammar/grammar.ebnf#L68-L70) of Molecule yet. To deal with the cases of importing other schemas, the codegen CLI also needs to 
> 
> 1. resolve all import schema items
> 2. remove the intersection with `prepend`
> 3. append them to the `prepend`
> 
> _Originally posted by @homura in https://github.com/ckb-js/lumos/issues/635#issuecomment-1961272941_

Note: 

The parser of Molecule written in JS doesn't support the `import` statement, therefore, it is better to handle the import statement separately. The instructions for handling the import feature could be

1. resolve the items from the `import` statement 
2. convert the exported items into ESM syntax and append these items into the [`prepend`](https://github.com/ckb-js/lumos/blob/develop/packages/molecule/src/codegen.ts#L13)
3. erase the `import` clause from the schema when calling the [codegen](https://github.com/ckb-js/lumos/blob/e34daa194c32c16dd8b5d5954e62326f15102593/packages/molecule/src/codegen.ts#L36) method
4. generate code recursively and name outputs the same with schema's

For example, now we have 2 schema files, `base` and `character`

```
// base.mol

array RGB [byte;3];
vector UTF8String <byte>;
```

```
// character.mol

import base;

table Character {
  hair_color: RGB, 
  name: UTF8String,
}
```

After the codegen, it should be

```ts
// base.ts
export const RGB = array(byte, 3);
export const UTF8String = vector(byte);
```

```ts
// character.ts

import { RGB, UTF8String } from './base'

export const Character = table({
  hair_color: RGB,
  name: UTF8String,
}, ['hair_color', 'name'])
```

coder-van:
# Description

generate TypeScript code recursively from mol type source file. Add config option output, when output = 0 (default), print TypeScript code only, and when output = 1, write to files.

Fixes #643

## Type of change

# How Has This Been Tested?

yarn run test

- [x] Test A
- [ ] Test B

# Checklist:


homura:
Thx for your pull request, it is inspiring for the feature design of multiple output targets. 

It would be better if

1. including some test cases
2. the `cli.ts` be cleaner to make it testable, it could be like this 

```ts
// these declarations are implemented separately from cli.ts

// the `AbsolutePath` is based on the schema file path instead of the file system's  
declare function resolveDependencies(entrySchema: code): Map<AbsolutePath, AbsolutePath[]>;

// similar to the `codegenReturnMore` in this PR, with some enhancement
declare function codegen(code: string): CodegenResult;

type CodegenResult = {
  code: string;
  elements: string[];
}

// helpers
declare function extractAndEraseImportClauses(code: string): string;
type AbsolutePath = string;

// so it can be more readable in the cli.ts
const dependencies = resolveDependencies(entrySource)
Object.entries(dependencies).forEach(([path, dependentPaths]) => {
  // ...
})

````

3. several typos, such as `bolck`, `recard`, `contine`
4. several uncommon abbreviations in comments, such as `ex: ...`

**Comment by coder-van** [comment]
OK, based on your suggestions, I will make optimizations

**Comment by coder-van** [comment]
1. AbsolutePath may be considered a system file path, so i thank RelativePath maybe more better;
2. function codegen is associated with the test script, I will modify it in subsequent work；
3. function loopCodegen need a sorted array from function resolveDependencies, Map structure does not provide correct ordering，so  it return `FileWithDependence[]` now.

If you have more suggestions, please continue to reply.

File: packages/molecule/src/resolve.ts [hunk 1]
```diff
@@ -0,0 +1,176 @@
...
+    relativePath: importPath,
+    dependencies: [],
+  };
+
+  const schema = fs.readFileSync(realPath, "utf-8");
+  if (!schema) {
+    return [cur];
+  }
+
+  const matched = schema.match(/.*import\s+"(.*)".*;/g);
```
homura, line 41: I found that the import statement definition should exclude the `""`, even though, there is something unexpected when a clause that includes a pattern of `import "";`, e.g.

```
vector string_import <byte>; // e.g. import "string";
```

maybe a [capturing group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Capturing_group) could be helpful here

packages/molecule/src/resolve.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (!resolved.has(mFilePath)) {
+      // mask this file has resolved
+      resolved.add(mFilePath);
+
+      const _dependencies = resolveDependencies(
+        mRelativePath,
+        baseDir,
+        resolved
+      );
+      dependencies.push(..._dependencies);
```
homura, line 73:
```suggestion
      dependencies.push(...resolveDependencies(
        mRelativePath,
        baseDir,
        resolved
      ));
```

It's better to avoid using variables that start with an underline as they often imply that the variable is private and mutable in a scope.

packages/molecule/src/resolve.ts
```diff
@@ -0,0 +1,176 @@
...
+
+export function extractAndEraseImportClauses(code: string): string {
+  const lines = code.split("\n");
+  const delImportLines = lines.filter((line: string) => {
+    return !line.trim().startsWith("import");
+  });
+  return delImportLines.join("\n");
+}
+
+function printOrWrite(resultMap: Map<string, ParseResult>, config: Config) {
```
homura, line 90: It is suggested to divide the function based on distinct output targets to make it more testable.

packages/molecule/src/resolve.ts
```diff
@@ -0,0 +1,176 @@
...
+        console.log(`write file ${tsName} finish`);
+      }
+    }
+  }
+}
+
+export function loopCodegen(config: Config): Map<string, ParseResult> {
+  const result: Map<string, ParseResult> = new Map();
+  const baseDir = path.dirname(config.schemaFile);
+  const relativePath = path.basename(config.schemaFile);
```
homura, line 123: I found that it's a `basename` rather than a `relativePath`, which could be misleading in this context

packages/molecule/src/resolve.ts
```diff
@@ -0,0 +1,176 @@
...
+export type FileWithDependence = {
+  relativePath: string;
+  dependencies: string[];
+};
+
+export function resolveDependencies(
+  importPath: RelativePath,
+  baseDir: string,
+  resolved: Set<RelativePath>
+): FileWithDependence[] {
```
homura, line 22: When defining a function with similar parameters, like `importPath`, `baseDir`, and `resolvedRelativePath`, it is recommended to comment them

homura:
Hi, @coder-van, your PR is inspiring, and I plan to make some improvements to it based on your work

coder-van:
> Hi, @coder-van, your PR is inspiring, and I plan to make some improvements to it based on your work

As for the details of the code, I may need some time to adjust it with the team. Thank you for your careful review, it is very helpful to me

packages/molecule/src/resolve.ts
coder-van: Good suggestion!

packages/molecule/src/resolve.ts
coder-van: Yes, good suggestion!

coder-van: Yes, indeed.

packages/molecule/src/resolve.ts
coder-van: Yes, that's better，agree with you.

packages/molecule/src/resolve.ts
coder-van: This is a detail I overlooked, it really has a big impact

packages/molecule/src/resolve.ts
coder-van: I will avoid this way of writing in the future

homura:
Hi, @coder-van, I've created a PR at #648 for addressing the issue #643. Thanks for your contribution again. I hope that this #648 PR will better express the optimizations I mentioned earlier.

coder-van:
I've read your code of PR #648, it's great, made me realize that in this project, higher code quality is needed, thank you.

coder-van:
After seeing your code, I suddenly realized that I wrote it in a simple way to complete the task, and you wrote it in order to achieve a complete solution, higher goals, and output better results. I saw several points from your output. What I praise:
1. Process all .mol in the directory (broader view);
2. Check for circular references (very rigorous);
3. More rigorous import checking, and processing of the content of comment blocks (I didn’t realize it at all);
4. Detailed test cases;

There may be a point that can be considered for optimization. Currently, after all files are read and the import statements are processed, the existence of dependent files is only checked when recursively building the path. Is it better to do it earlier?

homura:
> after all files are read and the import statements are processed, the existence of dependent files is only checked when recursively building the path. Is it better to do it earlier?

Exactly. We can detect the existence of the molecule file while building the dependency graph so that the process could exit earlier if any dependence is missed

homura:
@coder-van Thx for your PR that inspired me to create the #648 as an improvement. However, after reviewing your PR, I noticed a few issues, such as the lack of circular checking and test cases. I apologize for being nitpicky and closing it

PR #648, opened 2024-03-08
feat: support molecule import statement 

### LINKED ISSUE #643: Support the import statement of molecule for the codegen

homura:
# Description

Fixes #643 

This PR supports the import statement of Molecule but with a bit different from the design of #643. This PR supports codegen recursively for the pattern of `**/*.mol` in a directory instead of an entry file

## Type of change

# How Has This Been Tested?

- [x] unit test
- [x] test manually


packages/molecule/src/cli.ts
```diff
@@ -24,21 +41,83 @@ const fileConfig: Partial<Config> = (() => {
...
+  }
+
+  // remove the trailing slash for both windows and *nix
+  schemaDir = schemaDir.replace(/[\\/]+$/, "");
+
+  const namedSchemas = globSync(
+    path.join(schemaDir, "**/*.mol")
+  ).map<ProjectSchemaOptions>((molPath) => {
+    const prepend = (() => {
+      const match = config.prepend.match(importStatementRegex);
```
PainterPuppets, line 91: What is `prepend` used for in cli, will it allow multiple lines of import statements?

PainterPuppets (commented):
Looks great.

packages/molecule/src/cli.ts
homura: Currently, it is not supported to have custom schema items in separate files. It is recommended that you put all custom schema items into one file for unified management. Here is an example of how you can do this:

`number.ts`

```ts
export declare const Uint32Codec: AnyCodec;  
```

`string.ts`

```ts
export declare const StringCodec: AnyCodec;
```

If you want to override `Uint32` and `String` in all generated targets, you can create a `primitive.ts` file to export them from one entry point:

```json
{
  "prepend": "export { Uint32, String } from './primitive';"
}
```

Then, in `primitive.ts`, you can export `Uint32` and `String` as follows:

```ts
export { Uint32Codec as Uint32 } from './number'
export { StringCodec as String } from './string'
```

packages/molecule/src/topologySort.ts
coder-van, line 1: if a.mol -> b.mol -> (c.mol, d.mol) -> d.mol

in b.mol
import "c" 
import "d" 

in c.mol
import "d" 

if it will be treated as a cycle import?

homura, line 1: I suppose it is not a circular dependency if the file `d` doesn't import either 'b' or 'c'.

PainterPuppets (approved):

PR #653, opened 2024-03-18
feat: support joyid

### LINKED ISSUE #640: Support JoyID
It is recommended to provide an adapter to support JoyID for Lumos. JoyID now supports singing the `RawTransaction`.

```ts
import { lockScriptInfo } from '@ckb-lumos/joyid'

registerCustomLockScriptInfos(lockScriptInfo)
```

## Notes:

- https://docs.joyid.dev/guide/ckb/sign-raw-transaction

homura:
# Description

Fixes #640 

This PR brings a plugin for `registerCustomLockScriptInfos` to work with JoyID

```js
import { createJoyIDScriptInfo, getDefualtConfig } from "@ckb-lumos/joyid";
import { registerCustomLockScriptInfos } from "@ckb-lumos/lumos/common-scripts/common";
import { connect } from "@joyid/ckb";

// step 1. connect to JoyID
const connection = await connect();

// step 2. create JoyID script info, we can use the helper function getDefaultConfig to generate a default config
const joyIDScriptInfo = createJoyIDScriptInfo(
  connection,
  getDefaultConfig(false /* isMainnet */)
);

// step 3. register JoyID script info into Lumos
registerCustomLockScriptInfos([joyIDScriptInfo]);
```

This PR will target the branch `experimental-joyid` until `Lumos@0.22` is released. When this PR is merged, the NPM dist-tag 'experimental-joyid' will be used to pre-release the feature

## Type of change

# How Has This Been Tested?

- [x] Unit test
- [x] Manually test via example


examples/joyid/README.md
```diff
@@ -0,0 +1,12 @@
+# Lumos Works with JoyID
+
+[JoyID](https://joy.id/) is a universal account protocol. This example will show how to use Lumos to work with JoyID by a simple CKB transfer transaction.
```
Keith-CY:
```suggestion
[JoyID](https://joy.id/) is a universal account protocol. This example shows how to use Lumos to work with JoyID by a simple CKB transfer transaction.
```

packages/joyid/src/constants.ts
```diff
@@ -0,0 +1,82 @@
...
+    },
+    depType: "depGroup",
+  } as CellDep,
+};
+
+export const getJoyIDLockScript = (isMainnet = false): Script =>
+  isMainnet ? MainnetInfo.JoyIDLockScript : TestnetInfo.JoyIDLockScript;
+
+export const getJoyIDCellDep = (isMainnet = false): CellDep =>
+  isMainnet ? MainnetInfo.JoyIDLockDep : TestnetInfo.JoyIDLockDep;
```
Keith-CY, line 76: IMO `lock script` and `cell deps` can be returned together, e.g.

```ts
export const getJoyIdInfo = (isMainnet = false) => isMainnet ? MainnetInfo : TestnetInfo
```

By doing so, all necessary info will be available in a single call.

packages/joyid/src/script-info.ts
```diff
@@ -0,0 +1,269 @@
...
+  joyIdLockScriptTemplate: { codeHash: HexString; hashType: HashType };
+  cotaTypeScriptTemplate: { codeHash: HexString; hashType: HashType };
+};
+
+/* c8 ignore next 9*/
+export function getDefaultConfig(isMainnet: boolean): JoyIDScriptInfoConfig {
+  // https://github.com/nervina-labs/cota-sdk-js/blob/f80d04ea532d72cfe7410ea45af6dc583e140edf/README.md?plain=1#L46-L52
+  const aggregatorUrl = isMainnet
+    ? "https://cota.nervina.dev/mainnet-aggregator"
+    : "https://cota.nervina.dev/aggregator";
```
Keith-CY, line 103: What if passing the aggregator endpoint in for extension. When a boolean is in, using the default endpoint. When a string is in, using the custom endpoint.

packages/joyid/src/constants.ts
homura: this file is a copy of [JoyID constants](https://github.com/duanyytop/joyid-sdk-js/blob/4629f5158a5f54f00d765832d2d78fb20bb7c64b/src/constants/index.ts), so it is recommended to maintain its structure for easy syncing with the original

packages/joyid/src/script-info.ts
homura: Good point, I'll refactor it

packages/joyid/src/constants.ts
Keith-CY: Got it

packages/joyid/src/constants.ts
```diff
@@ -0,0 +1,82 @@
...
+    outPoint: {
+      txHash:
+        "0x636a786001f87cb615acfcf408be0f9a1f077001f0bbc75ca54eadfe7e221713",
+      index: "0x0",
+    },
+    depType: "depGroup",
+  } as CellDep,
+};
+
+const MainnetInfo = {
```
PainterPuppets, line 38: I see that these codes are not being used, and directly import from `joyid/ckb`, what's this here for?

PainterPuppets (approved):
Amazing work

There is only one small problem
The `packages/joyid` doesn't seem to have a `.gitignore` file.

packages/joyid/src/script-info.ts
homura: I realized that `getDefaultConfig` is a utility function for generating default configs by simply passing network or mainnet param. To customize the config, users can override the config by themself. I've commented it

```ts
const config = {
  ...getDefaultConfig(true),
  aggregator: new Aggregator('/path/to/my/aggregator')
}

createJoyIDScriptInfo(config)
```

packages/joyid/src/constants.ts
homura: the same reason here https://github.com/ckb-js/lumos/pull/653#discussion_r1529582163

packages/joyid/src/script-info.ts
Keith-CY: Got it

Keith-CY (approved):

PR #655, opened 2024-03-20
fix(ckb-indexer)!: wait for the next n blocks

homura:
# Description

The implementation of `waitForSync` is incorrect, although it should semantically wait for the next N blocks

## Type of change


PainterPuppets (approved):
LGTM

PR #674, opened 2024-04-23
feat(config-manager):add xudt config

sunchengzhu:
# Description

Adding XUDT config required for [the nervos docs site](https://github.com/nervosnetwork/docs.nervos.org/blob/0a8d6704924dd21977f1b3e108a27e0e1f3dd7f7/examples/xudt/lib.ts#L11)


Keith-CY:
xUDT config info comes from https://github.com/Magickbase/ckb-explorer-public-issues/issues/510

Keith-CY (approved):

PR #684, opened 2024-05-13
BREAKING CHANGE: drop all fs related support

homura:
## Description

When working with Lumos and CRA or other bundlers, polyfills are required for NodeJS modules. This PR removed all instances of `require("fs")` from `config-manager`, `hd`, and `hd-cache` modules

## Type of change

# How Has This Been Tested?

- [x] Unit test


Keith-CY (approved):

packages/hd-cache/src/index.ts
```diff
@@ -485,17 +487,21 @@ export function publicKeyToMultisigArgs(publicKey: HexString): HexString {
...
 
+  // hash160
+  // 0x prefix + 20 bytes hex = 2 + (20 * 2) = 42
+  const hash160HexLength = 42;
   const args = new utils.CKBHasher()
     .update(serialized)
     .digestHex()
-    .slice(0, 42);
+    .slice(0, hash160HexLength);
   return args;
```
twhy: These are optional
```ts
return utils.ckbHash(serialized).slice(0, hash160HexLength);
```
In another PR, adding a `type?: 'blade160'` param to `ckbHash()` 
```ts
function ckbHash(data: BytesLike, type?: 'blade160'): Hash {
  const LENGTHS = {
    blade160: 42
  }
  const hasher = new CKBHasher();
  hasher.update(bytes.bytify(data));
  return hasher.digestHex(0, LENGTHS[type] ?? Infinity);
}
```
After
```ts
return utils.ckbHash(serialized, 'blade160')
```

twhy (approved):

packages/hd-cache/src/index.ts
homura: Good point. It would be better if Lumos provided a function for `blake160` since `hash160` is common in CKB. As `CKBHasher` follows a general `Hasher` interface from NodeJS, a new function may be more suitable.

homura: I've added a new API `ckbHash160` to the `utils`. @twhy can you please review and check if the comment can be resolved

twhy (approved):

twhy (approved):

packages/hd-cache/src/index.ts
twhy: Sure. But it seems there's no `Resolve` button for comment.

PR #689, opened 2024-05-17
fix: correct error message for computeScriptHash

twhy:
# Description
This PR aims to fix error message "Expect type undefined ..." when calling `computeScriptHash` with invalid hex string.

image

## Type of change

# How Has This Been Tested?
Unit tests


.eslintrc.next.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
               2, // many .slice(2) since the '0x' prefix should be removed while calling 3rd-party library
               16, // toString(16)
               1000, // second to millisecond
+              0xff, // 255
```
homura: Perhaps we should ignore the ESlint compliance for the magick number on this [line](https://github.com/twhy/lumos/blob/41e517d8cf77b5cb6b858f19bbd297055e52c6bb/packages/codec/src/utils.ts#L82) since 0xff is not common in Lumos

twhy: OK. Done.

homura (approved):

PR #694, opened 2024-05-23
feat(helpers): model helpers for ckb object

homura:
# Description

This PR introduced a new helper `ModelHelper` to provide some common functions for CKB-related objects

### ModelHelpers

```ts
export type ModelHelper<Model, ModelLike = Model> = {
  create(modelLike: ModelLike): Model;
  equals(modelLike: ModelLike, modelR: ModelLike): boolean;
  hash(modelLike: ModelLike): Uint8Array;
  clone(model: ModelLike): Model;
};
```

```javascript
import { CellHelper } from "@ckb-lumos/helpers";

const cell = CellHelper.create({
  lock: "ckb1qzda0cr08m85hc8jlnfp3zer7xulejywt49kt2rr0vthywaa50xwsqgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq5m759c",
});

const _61CKB = 61 * 10 ** 8;
asserts(BI.from(cell.cellOutput.capacity).eq(_61CKB)); // 61 CKB
```

## Type of change

Tested?

- [x] Unit test


Keith-CY:
I didn't get the use case of `ModelHelper`. Just to instantiate a model easily with `hash`, `clone` and `equals` methods?

homura:
> I didn't get the use case of `ModelHelper`. Just to instantiate a model easily with `hash`, `clone` and `equals` methods?

There are two primary purposes of this PR:

- create CKB-related objects in a convenient way
- provide common methods to make the projects that are using Lumos more robust

## Convenience

In the past, if developers wanted to create a `Cell`, they had to use template code like this:

```js
const cell: Cell = {
  cellOutput: {
    lock: {...},
    capacity: '0x0'
  },
  data: '0x'
};

cell.cellOutput.capacity = minimalCapacityCompatible(cell).toHexString();
```

This process was heavy during transaction assembly. This PR introduces the concept of `ModelLike` to make the creation easier, simplifying it to:

```js
CellHelper.create({
  lock: 'ckt1...', // ScriptLike
  type: {
    codeHash: randomBytes(32), // BytesLike is acceptable
    hashType: 'type',
    args: Buffer.from([...]) // BytesLike is acceptable
  },
  data: Uint128.pack(111) // BytesLike is acceptable
} /* CellLike */);
```

## Robust

In the past, developers had multiple ways to interact with CKB-related objects, such as comparing two `Script`s. However, the code may not have been robust enough because they compared objects based on a string. For example, Neuron:
- [case 1](https://github.com/nervosnetwork/neuron/blob/4d1b3bbb3540a833c974b770f6758daf0efbb291/packages/neuron-wallet/src/services/cells.ts#L400-L420)
- [case 2](https://github.com/nervosnetwork/neuron/blob/4d1b3bbb3540a833c974b770f6758daf0efbb291/packages/neuron-wallet/src/services/tx/transaction-service.ts#L470-L471)

```js
if (
  codeHash === anotherCodeHash && 
  hashType === anotherHashType && 
  args === anotherArgs
) {
  // ...
}
```

This approach may cause unexpected behaviors, such as when the hex format string is not the same case, e.g. `0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee !== 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`. This is especially relevant when using Omnilock Ethereum authentication

Keith-CY:
> > I didn't get the use case of `ModelHelper`. Just to instantiate a model easily with `hash`, `clone` and `equals` methods?
> 
> There are two primary purposes of this PR:
> 
> * create CKB-related objects in a convenient way
> * provide common methods to make the projects that are using Lumos more robust
> 
> ## Convenience
> In the past, if developers wanted to create a `Cell`, they had to use template code like this:
> 
> ```js
> const cell: Cell = {
>   cellOutput: {
>     lock: {...},
>     capacity: '0x0'
>   },
>   data: '0x'
> };
> 
> cell.cellOutput.capacity = minimalCapacityCompatible(cell).toHexString();
> ```
> 
> This process was heavy during transaction assembly. This PR introduces the concept of `ModelLike` to make the creation easier, simplifying it to:
> 
> ```js
> CellHelper.create({
>   lock: 'ckt1...', // ScriptLike
>   type: {
>     codeHash: randomBytes(32), // BytesLike is acceptable
>     hashType: 'type',
>     args: Buffer.from([...]) // BytesLike is acceptable
>   },
>   data: Uint128.pack(111) // BytesLike is acceptable
> } /* CellLike */);
> ```
> 
> ## Robust
> In the past, developers had multiple ways to interact with CKB-related objects, such as comparing two `Script`s. However, the code may not have been robust enough because they compared objects based on a string. For example, Neuron:
> 
> * [case 1](https://github.com/nervosnetwork/neuron/blob/4d1b3bbb3540a833c974b770f6758daf0efbb291/packages/neuron-wallet/src/services/cells.ts#L400-L420)
> * [case 2](https://github.com/nervosnetwork/neuron/blob/4d1b3bbb3540a833c974b770f6758daf0efbb291/packages/neuron-wallet/src/services/tx/transaction-service.ts#L470-L471)
> 
> ```js
> if (
>   codeHash === anotherCodeHash && 
>   hashType === anotherHashType && 
>   args === anotherArgs
> ) {
>   // ...
> }
> ```
> 
> This approach may cause unexpected behaviors, such as when the hex format string is not the same case, e.g. `0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee !== 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`. This is especially relevant when using Omnilock Ethereum authentication

It's pretty OOP, so why not

```js
const cell = new Cell({
  codeHash: "hash"
})

// or

const cell = Cell.new({
  codeHash: 'hash'
})
```

and make the `hash`/`equals` a reader & getter as

```js
class Cell {
  static hash(cell) {
    if (cell instanceof Cell) return cell.hash
    return new Cell(cell).hash
  }

  static equals(a, b) {
    const [c, d] = [a, b].map(i => i instanceof Cell ? i : new Cell(i))
    return c.equals(d)
  }

  getter hash() {
    return this._hash()
  }

  equals (cell) {
    return this._equals(cell)
  }
}

const cellHash = new Cell().hash
Cell.equals(a, b)
new Cell(a).equals(b)
```

homura:
> It's pretty OOP, so why not

The `ModelHelpers` includes a set of common methods similar to OOP. However, the transaction assembling process of Lumos uses plain objects directly instead of classes. We can find a lot of code like this:

```js
const { injectCapacity } = common;

txSkeleton = await injectCapacity(txSkeleton, ...);
```

instead of 

```js
await txSkeleton.injectCapacity(...)
```

To keep a more consistent code style, `ModelHelpers` is designed as a utility instead of a superclass

OOP in JavaScript may cause some unexpected behavior like this:

```js
class Named {
  name = 'Alice';

  sayHi() {
    console.log('Hi, I am ' + this.name);
  }
}

const named = new Named();
const { sayHi } = named;

sayHi(); // Cannot read properties of undefined...
```

There were many object deconstructions used in Lumos or used with Lumos; therefore, a helper may be a more suitable for Lumos

Keith-CY:
> > It's pretty OOP, so why not
> 
> The `ModelHelpers` includes a set of common methods similar to OOP. However, the transaction assembling process of Lumos uses plain objects directly instead of classes. We can find a lot of code like this:
> 
> ```js
> const { injectCapacity } = common;
> 
> txSkeleton = await injectCapacity(txSkeleton, ...);
> ```
> 
> instead of
> 
> ```js
> await txSkeleton.injectCapacity(...)
> ```
> 
> To keep a more consistent code style, `ModelHelpers` is designed as a utility instead of a superclass
> 
> OOP in JavaScript may cause some unexpected behavior like this:
> 
> ```js
> class Named {
>   name = 'Alice';
> 
>   sayHi() {
>     console.log('Hi, I am ' + this.name);
>   }
> }
> 
> const named = new Named();
> const { sayHi } = named;
> 
> sayHi(); // Cannot read properties of undefined...
> ```
> 
> There were many object deconstructions used in Lumos or used with Lumos; therefore, a helper may be a more suitable for Lumos

It's counterintuitive to me because it looks like OOP but works in FP way. Maybe I'm misled by the name `CellHelper` which is usually a `Class` or `Enum`.

How about defining it as a `Class` with `static` methods.

Or simply name it `cellHelpers`

homura:
> Maybe I'm misled by the name `CellHelper` which is usually a `Class` or `Enum`.

Indeed, I hadn't realized that the first first capitalized camel naming implied that the reference should be a class instead of a plain object

> How about defining it as a `Class` with `static` methods.
> 
> Or simply name it `cellHelpers`

I'd prefer `cellHelper`

packages/helpers/README.md
```diff
@@ -35,31 +39,59 @@ const address = generateAddress({
...
+  equals(modelLike: ModelLike, modelR: ModelLike): boolean;
+  hash(modelLike: ModelLike): Uint8Array;
+  clone(model: ModelLike): Model;
+};
+```
+
+```javascript
+import { CellHelper } from "@ckb-lumos/helpers";
+
+const cell = CellHelper.create({
```
Keith-CY: `CellHelper` should be `cellHelper`

Keith-CY (approved):

PR #699, opened 2024-05-30
BREAKING CHANGE: disallow p2sh converting by default since risky

### LINKED ISSUE #680: Depreacte the support of Omnilock Bitcoin authentication's P2SH address
In the past, Omnilock supported Bitcoin, assuming all Bitcoin dApp users used UniSat wallet as their dApp entry. This meant that every P2SH address was assumed to be spendable with only one public key. However, this approach was not secure enough since P2SH outputs can be spent with any kind of scripts in Bitcoin.

## How to

To address this issue, the `decodeAddress` function in `omnilock/bitcoin` will be deprecated. Instead, a new function called `parseBtcAddressToPubkeyHash` will be created to provide more accurate results

This change will also affect the `createOmnilockScript` function, which will also be deprecated. Another function called `createSimplePubkeyBasedOmnilockScript`

homura:
# Description

Fixes #680

Since it is risky, this PR makes a breaking change to disallow p2sh address conversion by default.

## Type of change

# How Has This Been Tested?

- [x] Unit test


.changeset/green-pumpkins-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ckb-lumos/common-scripts": minor
+---
+
+**BREAKING CHANGE**: `createOmnilockScript` uses the `allow` option to restrict allowed btc addresses
```
Keith-CY:
```suggestion
**BREAKING CHANGE**: `createOmnilockScript` uses the `allows` option to restrict allowed btc addresses
```

packages/common-scripts/src/omnilock-bitcoin.ts
```diff
@@ -1,43 +1,56 @@
...
+): ArrayLike<number> {
+  const btcAddressFlagSize = 1;
+  const hashSize = 20;
 
-    // P2PKH
-    if (address.startsWith("1")) {
-      return bs58.decode(address).slice(1, 21);
-    }
+  if (isP2wpkhAddress(address) && allows.includes("P2WPKH")) {
+    return bech32.fromWords(bech32.decode(address).words.slice(1));
```
Keith-CY, line 25: `allows.includes("P2WPKH")` should be checked inside this branch and throw directly if `P2WPKH` is not allowed.

packages/common-scripts/src/omnilock-bitcoin.ts
```diff
@@ -1,43 +1,56 @@
...
  */
-export function decodeAddress(address: string): ArrayLike<number> {
-  try {
-    // Bech32
-    if (address.startsWith("bc1q")) {
-      return bech32.fromWords(bech32.decode(address).words.slice(1));
-    }
+export function decodeAddress(
+  address: string,
+  allows: SupportedBtcAddressType[]
```
Keith-CY: The default `allows` could be defined here.

homura: refactored

packages/common-scripts/src/omnilock-bitcoin.ts
homura: refactored

Keith-CY (approved):

PR #701, opened 2024-05-30
feat: add lumos crypto.randomBytes to replace node's

twhy:
# Description

1. Add `ckb-lumos/crypto` package which has a `randomBytes` method to replace node `crypto.randomBytes`.
2. export `ckb-lumos/crypto` in `ckb-lumos/lumos`

## Type of change

# How Has This Been Tested?

Unit tests


packages/codec/tests/bytes-like.test.ts
```diff
@@ -1,7 +1,7 @@
 import test from "ava";
 import { molecule, number } from "../src";
 import { Byte32 } from "../src/blockchain";
-import { randomBytes } from "crypto";
+import { randomBytes } from "../../crypto/src";
```
homura: Using the workspace feature offered by Pnpm to avoid the uninstalled package error. 

```suggestion
import { randomBytes } from "@ckb-lumos/crypto";
```

.github/.codecov.yml
```diff
@@ -12,6 +12,7 @@ coverage:
         target: auto
         threshold: 1%
         base: auto
+    patch: off # disable codecov/patch
```
homura: Is there any reason for adding the option here?

packages/crypto/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "name": "@ckb-lumos/crypto",
+  "version": "0.23.0-next.0",
+  "description": "Crypto utilities used in Lumos",
+  "author": "Tom Wang <tom.wanghaiyi@gmail.com@gmail.com>",
+  "homepage": "https://github.com/ckb-js/lumos#readme",
+  "license": "MIT",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "engines": {
+    "node": ">=20.0.0"
```
homura: Node v18 is still in [LTS until Apr 2025](https://nodejs.org/en/about/previous-releases), therefore, the maximum engine version here should be 18

packages/crypto/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  Object.defineProperty(bytes, "toString", {
+    value: (format?: "hex") => {
+      return format === "hex"
+        ? Array.from(bytes)
+            .map((byte) => byte.toString(16).padStart(2, "0"))
+            .join("")
+        : Uint8Array.prototype.toString.call(bytes);
+    },
+  });
```
homura: We should not use a nonstandard `Uint8Array` or it could confuse users and maintainers.

```js
methodStillNeedBuffer(Buffer.from(randomBytes(size)))
```

packages/hd/src/keystore.ts
```diff
@@ -94,8 +100,8 @@ export default class Keystore {
   // Create an empty keystore object that contains empty private key
   static createEmpty(): Keystore {
     const saltSize = 32;
-    const salt: Buffer = crypto.randomBytes(saltSize);
-    const iv: Buffer = crypto.randomBytes(16);
+    const salt: Buffer = Buffer.from(randomBytes(saltSize));
```
Keith-CY, line 103: Avoid using `Buffer` which is only available in Node.js.

`randomBytes` returns `Uint8Array` which can replace the `Buffer`

twhy: The plan is to replace `crypto` first, then replace `Buffer` with `Uint8Array` as described in  https://github.com/ckb-js/lumos/pull/645
image

packages/crypto/package.json
twhy: Updated as
```json
  "engines": {
    "node": ">=18.0.0"
  },
```

packages/crypto/src/index.ts
twhy: Removed.

packages/codec/tests/bytes-like.test.ts
twhy: Updated. 
Also added `"@ckb-lumos/crypto"` to `devDependecies` in `package.json` to avoid red wavy line in VSCode/IDE.

.github/.codecov.yml
twhy: `codecov/patch` complains about the 

` /* eslint-disable @typescript-eslint/no-magic-numbers */` comments added in this PR. 

**Is there a better way to do it other than disable it ?**

image

packages/hd/src/keystore.ts
Keith-CY: Got it

Keith-CY (approved):

.github/.codecov.yml
homura: this is because of some non-covered functions in the `mnemonic.ts` and this PR added some lines in `mnemonic.ts` that decreased the coverage rate. The report from Codecov is expected, so we should keep the patch coverage to understand the test coverage of a PR

twhy: OK. Enabled `codecov/patch` .

packages/crypto/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  if (size > MAX_BYTES) {
+    throw new RangeError(`size must be less than or equal to ${MAX_BYTES}`);
+  }
+
+  const bytes = new Uint8Array(size);
+  try {
+    crypto.getRandomValues(bytes); // Node.js 20+ and modern browsers
+  } catch {
+    /* eslint-disable @typescript-eslint/no-var-requires */
+    require("crypto").getRandomValues(bytes); // Node.js 18
```
homura: Consider using the `exports` field in `package.json` to avoid using the `require` API

twhy: How exactly ?

It seems `exports` field in `package.json` can not solve this problem.

`require("crypto").getRandomValues(...)` here targets only Node.js 18 because Web Crypto API 

`crypto.getRandomValues()` is not available in `global`

homura: sry, I misremembered the [browser](https://github.com/defunctzombie/package-browser-field-spec) keyword as being in `exports`, we should have used the `browser` keyword to export a different `crypto` runtime

twhy: The problem here is NOT about Browser or Node.js .

The problem here is about Node.js 18 or Node.js 20.

In modern browsers and Node.js 20, `crypto.getRandomValues()` is available in global namespace.

In Node.js 18, `crypto.getRandomValues()` is NOT in the global namespace,
so we have to `require('crypto')` to use it.

homura: For Nodejs-compatible runtimes, directly using the `require('crypto')` is fine. However, for a browser runtime or bundler, it doesn't always seem to work properly when using `require('package')` https://github.com/nervosnetwork/neuron/pull/3148

```js
// node.js

export { randomBytes } from 'crypto'

```

```js
// browser
export function randomBytes(size: number): Uint8Array {
  // ...
}
```

twhy: Added `crypto: false` in `browser` field in `package.json` to avoid issue like nervosnetwork/neuron#3148
```json
  "browser": {
    "crypto": false
  },
```
I've created a npm package https://www.npmjs.com/package/nocrypto to test the `browser` field behavior against webpack 5 and it works.

Browsers support `crypto.getRandomValues()` ten year ago so we don't have to worry about browsers.
https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API#browser_compatibility

For Node.js 18, `require('crypto')` is still necessary for now, but after May 2025, we can drop support for Node 18 and remove the `try {} catch {}` and `require('crypto')` .

We don't want code below because Node's `crypto.randomBytes` returns a `Buffer`.
```ts
// node.js
export { randomBytes } from 'crypto'
```

twhy: Updated to use `randomBytes` from `@noble/hashes` package

.github/.codecov.yml
```diff
@@ -12,6 +12,12 @@ coverage:
         target: auto
         threshold: 1%
         base: auto
+    patch:
+      default:
+        # basic settings
+        target: auto
+        threshold: 15%
+        base: auto
```
homura: the threshold is too large and it may cause problems, could you temporarily ignore the magic number in the file `mnemonic.ts` without modifying the config of the Codecov patch

```ts
// TODO comment for these magic numbers
/* eslint-disable @typescript-eslint/no-magic-numbers */
```

twhy: OK. Set to 1% now.

twhy:
@homura  Can this PR be merged into `develop` ?

homura:
> Can this PR be merged into `develop` ?

Not yet. We'll publish 0.23 first

twhy:
> > Can this PR be merged into `develop` ?
> 
> Not yet. We'll publish 0.23 first

Let's merge this and move forward.

homura:
> Let's merge this and move forward.

Maybe it's better to target this PR to the 'browser-compatibility' branch first until the node modules, such as 'buffer' and 'crypto', are fully dropped. Then, check if it works properly with Vite, CRA, and Next.js.

twhy:
The plan is to replace Node crypto first, then replace `Buffer` with `Uint8Array`

Now the first PR is done, and it's safe to merge it. The second is ready locally.

metioned in https://github.com/ckb-js/lumos/pull/645
image

twhy:
> > Let's merge this and move forward.
> 
> Maybe it's better to target this PR to the 'browser-compatibility' branch first until the node modules, such as 'buffer' and 'crypto', are fully dropped. Then, check if it works properly with Vite, CRA, and Next.js.

I will merge this PR into a newly created branch `browser` first.

PR #706, opened 2024-06-06
feat: move deprecate p2sh from 0.23 to 0.24

homura:
# Description

This PR reverts #682 because Lumos needs to release the JoyID support in 0.23, but the Omnilock P2SH auth mode is not completely supported yet

## Type of change

- [x] Other


Keith-CY (approved):

twhy (approved):

homura:
Added a changelog for version 0.23.0-next.1, and the changelog of the revert will be removed from the changeset automatic PR while release 0.23

PR #709, opened 2024-06-06
refactor: replace node crypto methods with noble methods

twhy:
# Description

This PR replaces Node crypto `createHash`, `createHmac`, `createCipheriv` ... with `@noble/hashes` and `@noble/ciphers`.

Once merged, there will be no `import {} from 'crypto'` or `import {} from 'node:crypto'` in Lumos packages.

## Type of change

# How Has This Been Tested?
Unit Tests


packages/hd/package.json
```diff
@@ -25,7 +25,6 @@
     "bn.js": "^5.1.3",
     "elliptic": "^6.5.4",
     "scrypt-js": "^3.0.1",
```
homura, line 27: The `scrypt-js` library relies on the `buffer`. Is it possible to replace it with the `noble/cipher`'s scrypt?

packages/hd/src/keychain.ts
```diff
@@ -1,6 +1,7 @@
-import crypto from "crypto";
-import { ec as EC } from "elliptic";
+/* eslint-disable @typescript-eslint/no-magic-numbers */
```
homura, line 1: does the linting magic number affect multiple lines of `keychain.ts`?

packages/hd/src/mnemonic/index.ts
```diff
@@ -50,37 +55,21 @@ function salt(password = ""): string {
...
-    } catch (error) {
-      reject(error);
-    }
-  });
+  const mnemonicBuffer = Buffer.from(mnemonic.normalize("NFKD"), "utf8");
+  const saltBuffer = Buffer.from(salt(password.normalize("NFKD")), "utf8");
+  return pbkdf2Async(sha512, mnemonicBuffer, saltBuffer, {
+    c: PBKDF2_ROUNDS,
+    dkLen: KEY_LEN,
+  }).then(Buffer.from);
```
homura, line 72: these lines aren't covered. could you please add tests for them? 

vector and cases can be found in the bip39 repo https://github.com/bitcoinjs/bip39/blob/4fce199e73f7ab68bdf9f7382dc10f7e1adaa499/test/index.js

packages/crypto/src/index.ts
```diff
@@ -1 +1,9 @@
+export { ctr } from "@noble/ciphers/aes";
+export { hmac } from "@noble/hashes/hmac";
+export { sha256 } from "@noble/hashes/sha256";
+export { sha512 } from "@noble/hashes/sha512";
+export { keccak_256 } from "@noble/hashes/sha3";
```
homura:
```suggestion
export { keccak_256 as keccak256 } from "@noble/hashes/sha3";
```

we should avoid mixing snake_case with camelCase in a project if possible

packages/hd/src/keychain.ts
```diff
@@ -1,6 +1,7 @@
...
 import BN from "bn.js";
+import { ec as EC } from "elliptic";
```
homura, line 3: can the library 'elliptic' be replaced with the `noble/curve` or `noble/secp256k1`?

packages/hd/src/keystore.ts
```diff
@@ -151,19 +144,17 @@ export default class Keystore {
...
-      cipher.update(
-        Buffer.from(extendedPrivateKey.serialize().slice(hexPrefixSize), "hex")
-      ),
-      cipher.final(),
-    ]);
+    // DO NOT remove the Uint8Array.from call below.
+    // Without calling Uint8Array.from to make a copy of iv,
+    // iv will be set to 0000...00000 after calling cipher.encrypt(plaintext)
+    // and decrypting the ciphertext will fail
+    /* eslint-disable @typescript-eslint/no-magic-numbers */
```
homura, line 158: i found that the number 2 is common in Lumos since the `str.slice(2)` is used often, so i'd suggest adding the 2 to ignore list to avoid the disable annotation

packages/hd/src/keystore.ts
```diff
@@ -230,18 +217,15 @@ export default class Keystore {
...
+    // https://github.com/ethereumjs/ethereumjs-wallet/blob/d57582443fbac2b63956e6d5c4193aa8ce925b3d/src/index.ts#L615-L617
+    // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+    const hash = keccak_256(
+      Buffer.concat([derivedKey.subarray(16, 32), ciphertext])
     );
+    return Buffer.from(hash).toString("hex");
   }
 
-  static scryptOptions(kdfparams: KdfParams): ScryptOptions {
+  static scryptOptions(kdfparams: KdfParams): ScryptOpts {
```
homura: Is `ScryptOpts` the same as `ScryptOptions`? It would be better to copy it from "crypto.d.ts" to ensure the method is not changed

packages/hd/package.json
twhy: Yes. This will be done in next PR which I will replace `Buffer` with `Uint8Array`
This PR is more focus on replacing node crypto.

packages/hd/src/keychain.ts
twhy: `keychain.ts` has 7 magic numbers so this disable `no-magic-numbers` rule for the whole file 
just like `mnemonic/index.ts`.

If disable the rule line by line, it will break `codecov/patch`

packages/hd/src/mnemonic/index.ts
twhy: Both `mnemonicToSeed` and `mnemonicToSeedSync` are covered by tests.
https://github.com/ckb-js/lumos/blob/develop/packages/hd/tests/mnemonic/index.test.ts#L42

packages/crypto/src/index.ts
twhy: Done.

packages/hd/src/keychain.ts
twhy: Yes. This can be done. 
1. Replace node crypto.
2. Replace `Buffer` with `Uint8Array`
3. Optimize bundle package size.    `elliptic` will be replaced in this step.

packages/hd/src/keystore.ts
twhy: Added (back).
It was there before. 

In `.eslintrc.next.js`
```ts
        "@typescript-eslint/no-magic-numbers": [
          "error",
          {
            ignoreEnums: true,
            ignoreArrayIndexes: true,
            ignoreClassFieldInitialValues: true,
            ignoreDefaultValues: true,
            ignore: [
              -1, // index -1 is not found
              0, // first element of an array
              1, // common for i + 1 in a loop
              2, // many .slice(2) since the '0x' prefix should be removed while calling 3rd-party library
              16, // toString(16)
              1000, // second to millisecond
            ],
          },
        ],
```

packages/hd/src/keystore.ts
twhy: Both `ScryptOptions` and `ScryptOpts` provide inaccurate return type.
Based on the method body, the most accurate return type should be just no explicit return type.
```ts
  static scryptOptions(kdfparams: KdfParams) {
    return {
      N: kdfparams.n,
      r: kdfparams.r,
      p: kdfparams.p,
      // eslint-disable-next-line @typescript-eslint/no-magic-numbers
      maxmem: 128 * (kdfparams.n + kdfparams.p + 2) * kdfparams.r,
    };
  }
```
```ts
// noble scrypt.d.ts
export type ScryptOpts = {
    N: number;
    r: number;
    p: number;
    dkLen?: number;
    asyncTick?: number;
    maxmem?: number;
    onProgress?: (progress: number) => void;
};

// node crypto.d.ts
interface ScryptOptions {
    cost?: number | undefined;
    blockSize?: number | undefined;
    parallelization?: number | undefined;
    N?: number | undefined;
    r?: number | undefined;
    p?: number | undefined;
    maxmem?: number | undefined;
}
```

homura: I'm not sure I understand the meaning of "inaccurate return type" 

The original method returns the type of:

```ts
interface InferReturnType {
      N: number,
      r: number,
      p: number,
      maxmem: number,
}
```

This type is compatible with the definition `ScryptOptions` in `crypto.d.ts`. So, what does "inaccurate type" refer to?

Changing `ScryptOptions` to `ScryptOpts` is a breaking change that could lead to unexpected results for users who rely on the API directly.

.changeset/gorgeous-yaks-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ckb-lumos/crypto": minor
+"@ckb-lumos/hd": minor
+---
+
+replace node crypto methods with @noble/hashes and @noble/ciphers
```
homura:
```suggestion
refactor: replace node crypto methods with @noble/hashes and @noble/ciphers
```

packages/hd/src/keystore.ts
twhy: I simply removed the return type of this method in latest update.

This is how the method looks like now.
```ts
  static scryptOptions(kdfparams: KdfParams) {
    return {
      N: kdfparams.n,
      r: kdfparams.r,
      p: kdfparams.p,
      // eslint-disable-next-line @typescript-eslint/no-magic-numbers
      maxmem: 128 * (kdfparams.n + kdfparams.p + 2) * kdfparams.r,
    };
  }
```


This is the accurate return type
```ts
interface InferReturnType {
      N: number,
      r: number,
      p: number,
      maxmem: number,
}
```
Both `ScryptOptions` and `ScryptOpts` are just OK, but inaccurate.

homura: Or can you define the return type explicitly?

```diff
+type ScryptOptions = {
+  N: number;
+  r: number;
+  p: number;
+  maxmem: number;
+}
```

Also, remove the `ScryptOptions` from import, as it is compatible with both `ScryptOpts` and `crypto.ScryptOptions`

twhy: Done.

.changeset/gorgeous-yaks-notice.md
twhy: Added.

homura (approved):

Keith-CY (approved):

PR #710, opened 2024-06-11
BREAKING CHANGE: deploy defaults to data1

homura:
# Description

This PR changes the default `hash_type` of `generateDeployWithDataTx` to `data1` since `data2` is not active on the testnet or mainnet yet

## Type of change


Keith-CY:
This update should be treated as a breaking change because the default value changes.

Keith-CY (approved):

PR #711, opened 2024-06-11
BREAKING CHANGE: remove isDeepEqual function

homura:
# Description

This PR removes the function `isDeepEqual` since it is not used in other packages to reduce the package size

## Type of change


twhy (approved):

Keith-CY:
It's not a breaking change because `isDeepEqual` has not been exported to users.

Keith-CY:
> It's not a breaking change because `isDeepEqual` has not been exported to users.

Fix myself, it's been exported

Keith-CY (approved):

PR #716, opened 2024-06-21
docs: add docs for v0.30

twhy:
# Description

Add docs for v0.30

## Type of change

# How Has This Been Tested?
None


website/docs/migrations/migrate-to-v0.30.md
```diff
@@ -0,0 +1,74 @@
...
+
+You can still pass in `Buffer` because `Buffer` is a subclass of `Uint8Array`, but using `Uint8Array` is recomended.
+
+`@ckb-lumos/hd/KeyChain`
+
+```diff
+export default class Keychain {
+  ...
+- constructor(privateKey: Buffer, chainCode: Buffer)
++ constructor(privateKey: Uint8Array, chainCode: Uint8Array)
```
homura, line 52: when `Buffer` is present in the parameter, there won't be any breaking change since it is [extended from `Uint8Array`](https://github.com/nodejs/node/blob/a0cb507ea67121392082795943a3dc00e5317b5d/doc/api/buffer.md#buffer). Thus, it's advisable to announce that these changes allow the APIs to use `Uint8Array`, and switching to `Uint8Array` is not mandatory

twhy: image
Yes. I did that.

homura (approved):

PR #718, opened 2024-06-24
feat(common-scripts): support eth displaying for omnilock

### LINKED ISSUE #717: Got error when transfer ckb using omni_lock address
**Describe the bug**
Got error when transfer CKB using omni_lock address :
`Unsupported flag: 18, please check if the script.argsis expected`
It only appear at version 0.23.0, I downgrade lumos to resolve it.

**version**
@ckb-lumos/lumos : 0.23.0

**To Reproduce**
```
 const txSkeleton = await commons.common.transfer(
            _txSkeleton,
            [<omni_lock_address>],
            to,
            BigInt(amount),
        )
```

**Expected behavior**
Not throwing error

**Screenshots**
微信图片_20240621133329

homura:
# Description

Fixes #717

This PR supports the [`IdentityFlagsEthereumDisplaying`](https://github.com/cryptape/omnilock/blob/cd764d7133ec4e6b192fac4b93fc0596ef5b71f6/c/ckb_identity.h#L72) auth mode for Omnilock

## Quick Start

```ts
import { commons } from '@ckb-lumos/lumos'

const signature = await commons.omnilock.ethereumDisplaying.signMessage(
  provider.selectedAddress,
  txSkeleton.get("signingEntries").get(0)!.message,
  window.ethereum
);

const signedWitness = bytes.hexify(
    blockchain.WitnessArgs.pack({
      lock: OmnilockWitnessLock.pack({ signature }),
    })
  );

txSkeleton = txSkeleton.update('witnesses', inputIndexOfTheOmnilock)

```

## Type of change

# How Has This Been Tested?

- [x] Unit test


packages/common-scripts/src/omnilock-ethereum-displaying.ts
```diff
@@ -0,0 +1,51 @@
+import { BytesLike, bytes } from "@ckb-lumos/codec";
+import { hexify } from "@ckb-lumos/codec/lib/bytes";
+
+const COMMON_PREFIX = "CKB transaction: 0x";
+
+export interface Provider {
+  request: {
+    (payload: {
+      method: "personal_sign";
+      params: [string /*from*/, string /*message*/];
```
Keith-CY: It can be defined as 
```ts
params: [from: string, message: string]
```

Ref: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements

packages/common-scripts/src/omnilock-ethereum-displaying.ts
```diff
@@ -0,0 +1,51 @@
...
+      window.ethereum
+    ) {
+      return window.ethereum as Provider;
+    }
+
+    throw new Error(
+      "No provider found, make sure you have installed MetaMask or the other EIP1193 compatible wallet"
+    );
+    /* c8 ignore stop */
+  })();
```
Keith-CY: Why not assign the provider directly? E.g.

```ts
let internal = provider
if (!internal) {
  ...
}
```

packages/common-scripts/src/omnilock.ts
```diff
@@ -49,6 +49,7 @@ export type OmnilockInfo = {
 export type OmnilockAuth =
   | IdentityCkb
   | IdentityEthereum
+  | IdentityEthereumDisplaying
```
Keith-CY, line 52: What does `Identity Ethereum Displaying` mean

homura: https://github.com/cryptape/omnilock/blob/cd764d7133ec4e6b192fac4b93fc0596ef5b71f6/c/ckb_identity.h#L72

I used the same name as the source from Omnilock

packages/common-scripts/src/omnilock-ethereum-displaying.ts
homura: My personal coding habit in JavaScript: I aim to avoid using the `let` keyword whenever possible

packages/common-scripts/src/omnilock-ethereum-displaying.ts
homura: refactored

packages/common-scripts/src/omnilock-ethereum-displaying.ts
Keith-CY: The `let` can be avoided by
```
const internal = provider ?? globalThis?.['ethereum']
```

packages/common-scripts/src/omnilock.ts
Keith-CY: Got it

packages/common-scripts/src/omnilock-ethereum-displaying.ts
homura: good point, i've simplified the code

Keith-CY (approved):

PR #729, opened 2024-07-11
feat(lumos): export more missing modules

homura:
# Description

This PR 

- Exporting some missing common modules, such as `TransactionCollector` from `@ckb-lumos/lumos`
- Splitting sub-packages from `@ckb-lumos/lumos` so that SDK authors based on Lumos can import them on demand
- Adding a bundler playground to estimate the bundle size

To make the app bundle smaller, developers could import from sub-packages on demand like the following demonstration:

```ts
import { type Script } from '@ckb-lumos/lumos/base'

const script:Script = {
  codeHash: '0x...',
  hashType: 'type',
  args: '0x...',
}

```


packages/hd/package.json
```diff
@@ -23,7 +23,7 @@
     "@ckb-lumos/bi": "0.24.0-next.0",
     "@ckb-lumos/codec": "0.24.0-next.0",
     "@ckb-lumos/crypto": "0.24.0-next.0",
-    "bn.js": "^5.1.3",
+    "bn.js": "^4.12.0",
```
Keith-CY, line 26: Why is `bn.js` downgraded?

Keith-CY (approved):

packages/hd/package.json
homura: The `elliptic` library in use depends on the `bn.js@4.12.0`. To prevent multiple versions of `bn.js`, this PR downgrades `bn.js` from 5.1 to 4.12

Keith-CY: Got it

PR #738, opened 2024-08-06
feat(molecule): support the latest mol grammar

homura:
# Description

The previous version of `@ckb-lumos/molecule` does not fully support all molecule syntax, such as `comment` starting with `#`, `comment`s in table or struct declaration blocks, `import` syntax, `syntax` version, newline breaker `\r\n` in Windows, and more. This PR refactors the parser to support the latest Molecule syntax.

- fully support the test version syntax of Molecule 0.8
- replaced the parser with Peggy, [a successor of peg.js](https://github.com/pegjs/pegjs/issues/675), so the output parser does not require any runtime

## Type of change

# How Has This Been Tested?

- [x] Unit test


packages/molecule/src/grammar/grammar.js
```diff
@@ -0,0 +1,1980 @@
+// @generated by Peggy 4.0.3.
```
Keith-CY: This file should not be included in git, but generated when the package is being published.

homura: The codegen file is removed and ignored. It will be generated when running `npm run build`

homura:
Now developers can paste their `.mol` with the latest Molecular syntax in the online Molecular parser.

Here is an example from [UTXO Stack](https://github.com/ckb-cell/utxo-stack-sdk/blob/develop/packages/leap/src/request/request.spec.ts) and its corresponding schema file [`leap.mol`](https://github.com/ckb-cell/utxo-stack-sdk/blob/develop/packages/leap/src/molecule/schemas/leap.mol)

image

Keith-CY (approved):

